PR #9, opened 2019-07-10
Add upgrade

jayfjin:
Add upgrade capability for cdap operator.

Upgrade request is made by passing in desired new image in CR fields queuedImage and queuedUserInterfaceImage. Operator notices these fields, and starts up a pre-upgrade job. Operator waits until this job is finished, and then updates the image. User should query for state not upgrade failed, and ready to denote the upgrade finished successfully in status.meta.components.


config/crds/cdap_v1alpha1_cdapmaster.yaml
```diff
@@ -197,6 +197,12 @@ spec:
                     the service.
                   type: string
               type: object
+            queuedImage:
```
chtyim: Why need a new field? How does it relate to the image field? After the upgrade will this goes retained? Will the image field get changed?

For other controller, typically upgrade is done through updating of the image field

config/crds/cdap_v1alpha1_cdapmaster.yaml
```diff
@@ -235,6 +241,10 @@ spec:
             serviceAccountName:
               description: The service account for all the service pods.
               type: string
+            upgradeJobStartTimeNano:
+              description: Upgrade Job start time in nano, for internal use only
```
chtyim: How is it being used and who is using it? For internal states used by the controller, they should be stored in the status section. The spec section is for user to define the specification, which shouldn’t have any internal stuff

config/crds/cdap_v1alpha1_cdapmaster.yaml
jayfjin: The image field gets changed by the operator once the job finishes. Reason for having this as a separate field is that the clh will query the CR, so if we update the image then there's a race condition between querying the CR and having the operator set the image back to the old version while waiting for the job to run.

chtyim: Anything that is being changed by the operator and query by external entities should be under status, not under spec.

jayfjin: Done

pkg/apis/cdap/v1alpha1/cdapmaster_types.go
```diff
@@ -133,10 +133,21 @@ type UserInterfaceSpec struct {
 	CDAPExternalServiceSpec `json:",inline"`
 }
 
+// UpgradeJobSpec defines the specification for the upgrade job
+type UpgradeJobSpec struct {
```
chtyim: This is not a public type that is used by the user, right? Seems like it should be just a local struct in the controller

pkg/apis/cdap/v1alpha1/cdapmaster_types.go
```diff
@@ -133,10 +133,21 @@ type UserInterfaceSpec struct {
...
+	JobName  string `json:"jobName,omitempty"`
+	HostName string `json:"hostName,omitempty"`
+}
+
 // CDAPMasterStatus defines the observed state of CDAPMaster
 type CDAPMasterStatus struct {
-	status.Meta          `json:",inline"`
+	status.Meta `json:",inline"`
 	status.ComponentMeta `json:",inline"`
+	IsUpgrading                bool   `json:"isUpgrading"`
```
chtyim: Better use the `Condition` field in the `status.Meta` to report upgrade status. This is the common way in K8s. Upgrade failure can also be reported through that field.

BTW, when naming boolean field, name it without `is` prefix.

pkg/apis/cdap/v1alpha1/cdapmaster_types.go
```diff
@@ -133,10 +133,21 @@ type UserInterfaceSpec struct {
...
+	ExistingImage              string `json:"existingImage"`
```
chtyim: It is not very clear what existing means. Do you mean old image or current image?

pkg/controller/cdapmaster/cdapmaster_controller.go
```diff
@@ -389,15 +415,79 @@ func (b *Base) Observables(rsrc interface{}, labels map[string]string) []reconci
 
 // Objects for Messaging service
 func (s *Messaging) Objects(rsrc interface{}, rsrclabels map[string]string, observed, dependent, aggregated []reconciler.Object) ([]reconciler.Object, error) {
-	r := rsrc.(*alpha1.CDAPMaster)
-	return getStatefulServiceResources(&r.Spec.Messaging.CDAPStatefulServiceSpec, rsrc, rsrclabels, alpha1.ServiceMessaging)
+	master := rsrc.(*alpha1.CDAPMaster)
```
chtyim: Why this logic in the object handler for the Messaging service? Shouldn't it be done in the Base object?

pkg/controller/cdapmaster/cdapmaster_controller.go
```diff
@@ -389,15 +415,79 @@ func (b *Base) Observables(rsrc interface{}, labels map[string]string) []reconci
...
+	for _, item := range reconciler.ObjectsByType(observed, k8s.Type) {
+		if job, ok := item.Obj.(*k8s.Object).Obj.(*batchv1.Job); ok {
+			if master.Status.UpgradeJobStartTimeNano != 0 && strings.Contains(job.Name, strconv.FormatInt(master.Status.UpgradeJobStartTimeNano, 10)) {
+				upgradeJob = job
+				break
+			}
+		}
+	}
+	var upgradeResources []reconciler.Object
+	if upgradeJob == nil && master.Status.UpgradeJobStartTimeNano == 0 &&
```
chtyim: This function is kind of very big with multiple upgrade state handling inside. Please refactor them into multiple functions to make the intentions clear (launchJob, check Job, on Job finish).

pkg/controller/cdapmaster/cdapmaster_controller.go
```diff
@@ -389,15 +415,79 @@ func (b *Base) Observables(rsrc interface{}, labels map[string]string) []reconci
...
+				HostName: getConfigName(master, "router"),
+			}, rsrclabels)
+		if err != nil {
+			return nil, err
+		}
+	} else if upgradeJob != nil && upgradeJob.Status.Succeeded > 0 {
+		// job finished
+		master.Status.UpgradeJobStartTimeNano = 0
+		master.Status.IsUpgrading = false
+	} else if upgradeJob != nil && (time.Now().Add(-20*time.Minute).UnixNano() > master.Status.UpgradeJobStartTimeNano || upgradeJob.Status.Failed >= 4) {
```
chtyim: I think the controller shouldn't automatically change the Spec. The spec is the desire state as declared by the user, the controller will try its best to make current state to converge to the desire state. The upgrade Job itself can fail due to various reason (including timeout, but the Job should exit itself with fail state, instead of getting killed externally). The controller should reflect failure in the Status and the user/caller can monitor the status and decide what to do if it fails.

templates/upgrade-job.yaml
```diff
@@ -0,0 +1,27 @@
...
+spec:
+  template:
+    metadata:
+    {{if .Labels}}
+    labels:
+        {{range $k,$v := .Labels}}
+          {{$k}}: '{{$v}}'
+          {{end}}
+    {{end}}
+    spec:
```
chtyim, line 34: I think we need to specify the termination and cleanup as per https://kubernetes.io/docs/concepts/workloads/controllers/jobs-run-to-completion/#job-termination-and-cleanup

jayfjin:
Addressed comments, please take another look!

config/crds/cdap_v1alpha1_cdapmaster.yaml
```diff
@@ -370,12 +370,25 @@ spec:
                 - status
                 type: object
               type: array
+            oldImage:
```
chtyim: What is this for?

config/crds/cdap_v1alpha1_cdapmaster.yaml
```diff
@@ -370,12 +370,25 @@ spec:
...
+              description: Docker image of CDAP backend of the previous iteration.
+              type: string
+            oldUserInterfaceImage:
+              description:  Docker image of CDAP UI of the previous iteration.
+              type: string
+            useOldImage:
```
chtyim: What is this for?

pkg/controller/cdapmaster/cdapmaster_controller.go
```diff
@@ -20,27 +20,42 @@ import (
...
+	templateDir              = "templates/"
+	deploymentTemplate       = "cdap-deployment.yaml"
+	uiDeploymentTemplate     = "cdap-ui-deployment.yaml"
+	statefulSetTemplate      = "cdap-sts.yaml"
+	serviceTemplate          = "cdap-service.yaml"
+	upgradeJobTemplate       = "upgrade-job.yaml"
+
+	upgrading                = "upgrading"
+	upgradeFailed            = "upgrade-failed"
+	upgradeSucceededMessage  = "upgrade-succeeded"
```
chtyim: Should this be `upgradeSucceeded`?

config/crds/cdap_v1alpha1_cdapmaster.yaml
jayfjin: It's used by the templates to keep the image from changing during upgrade until the job is completed.

config/crds/cdap_v1alpha1_cdapmaster.yaml
jayfjin: Used by the templates to condition on either using oldImage or the image in the spec. We don't want to use the image in the spec until the pre upgrade job is done

pkg/controller/cdapmaster/cdapmaster_controller.go
jayfjin: No, it's just the message that we set upgrading to when it finishes. There are only two cond types, upgrading and upgradeFailed. On success, both are in state false, which is what the CLH should query for.

pkg/controller/cdapmaster/cdapmaster_controller.go
```diff
@@ -376,6 +471,49 @@ func (b *Base) Objects(rsrc interface{}, rsrclabels map[string]string, observed,
 		}
 	}
 
+	var upgradeJob *batchv1.Job
+	for _, item := range reconciler.ObjectsByType(observed, k8s.Type) {
```
barney-s: You can use: k8s/GetItem()

https://github.com/cdapio/cdap-operator/blob/fe748c2ac36b4de7b784a8ad8360fb9c6d469de2/vendor/sigs.k8s.io/controller-reconciler/pkg/reconciler/manager/k8s/manager.go#L358

barney-s (commented):
General feedback:
1. Upgrade is a common pattern than has not been addressed in the controller-reconciler library. We can add it there to be useful for all controllers.
2. I believe you are using .status.oldimage and .status.useoldimage to track state and progress of upgrade and avoid mutating the sts to new image. Alternatively you can replace the sts in Objects()'s resources from observed which is the old instance. Then all you need to track is .status.upgradingtoImage.

config/crds/cdap_v1alpha1_cdapmaster.yaml
```diff
@@ -376,6 +376,19 @@ spec:
                 by the API Server.
               format: int64
               type: integer
+            oldImage:
```
chtyim: Maybe better call it `upgradeFromImage` since we don't have a `newImage`.

config/crds/cdap_v1alpha1_cdapmaster.yaml
```diff
@@ -376,6 +376,19 @@ spec:
...
+              description: OldImage is the Docker image of CDAP backend of the previous iteration.
+              type: string
+            oldUserInterfaceImage:
+              description: OldUserInterfaceImage is the Docker image of CDAP UI of the previous iteration.
+              type: string
+            upgradeJobStartTimeNano:
+              description: The start time in nano of the upgrade job.
+              format: int64
+              type: integer
+            useOldImage:
```
chtyim: Why this is needed? If the upgrade job is still running or failed, the the controller should be using old image to populate all yaml templates, right?

pkg/controller/cdapmaster/cdapmaster_controller.go
```diff
@@ -20,27 +20,41 @@ import (
...
+	javaMinHeapRatio         = float64(0.6)
+	javaReservedNonHeap      = int64(768 * 1024 * 1024)
+	templateDir              = "templates/"
+	deploymentTemplate       = "cdap-deployment.yaml"
+	uiDeploymentTemplate     = "cdap-ui-deployment.yaml"
+	statefulSetTemplate      = "cdap-sts.yaml"
+	serviceTemplate          = "cdap-service.yaml"
+	upgradeJobTemplate       = "upgrade-job.yaml"
+
+	preUpgrading             = "pre-upgrading"
```
chtyim: preUpgrade

jayfjin:
Thanks Barni. Addressed 2. For 1, I think we should keep this as is for now for timing reasons, but happy to work with you after we get upgrade done to make this generalizable, as currently this is custom tailored for our needs.

pkg/controller/cdapmaster/cdapmaster_controller.go
```diff
@@ -20,27 +20,39 @@ import (
...
+
+	upgradeFailed            = "upgrade-failed"
+	upgradeStartMessage      = "Upgrade started, received updated CR."
+	upgradeFailedInitMessage = "Failed to create job, upgrade failed."
+	upgradeJobFailedMessage  = "Upgrade job failed."
+	upgradeResetMessage      = "Upgrade spec reset."
+	upgradeFailureLimit      = 4
 )
 
 // Add creates a new ESCluster Controller and adds it to the Manager with default RBAC. The Manager will set fields on the Controller
```
chtyim: BTW, the comment seems wrong. There is no ESCluster, right?

pkg/controller/cdapmaster/cdapmaster_controller.go
```diff
@@ -236,11 +248,41 @@ type externalServiceValue struct {
...
 }
 
+// Struct containing data for templatization using UpgradeJobSpec
+type upgradeValue struct {
+	templateBaseValue
+	Job *UpgradeJobSpec
+}
+
+// UpgradeJobSpec defines the specification for the upgrade job
+type UpgradeJobSpec struct {
```
chtyim: Does this type needs to be public?

pkg/controller/cdapmaster/cdapmaster_controller.go
```diff
@@ -236,11 +248,41 @@ type externalServiceValue struct {
...
+	BackoffLimit int32  `json:"backoffLimit, omitempty"`
+}
+
 // Returns the config map name for the given configuration type
 func getConfigName(r *alpha1.CDAPMaster, confType string) string {
 	return fmt.Sprintf("cdap-%s-%s", r.Name, confType)
 }
 
+// Gets the set of resources for the pre-upgrade job.
+func getUpgradeResources(s *UpgradeJobSpec, rsrclabels map[string]string) ([]reconciler.Object, error) {
```
chtyim: Should be called `getPreUpgradeResources`.

pkg/controller/cdapmaster/cdapmaster_controller.go
```diff
@@ -236,11 +248,41 @@ type externalServiceValue struct {
...
+	ngdata := &upgradeValue{Job: s}
+	ngdata.Labels = rsrclabels
+	item, err := k8s.ObjectFromFile(templateDir+upgradeJobTemplate, ngdata, &batchv1.JobList{})
```
chtyim: Prefer to have spaces around the `+`.

pkg/controller/cdapmaster/cdapmaster_controller.go
```diff
@@ -236,11 +248,41 @@ type externalServiceValue struct {
...
+	if err != nil {
+		return nil, err
+	}
+	return append([]reconciler.Object{}, *item), nil
+}
+
+// Gets the name of the upgrade job based on start time.
+func getUpgradeJobName(startTime int64) string {
+	return "upgrade-job" + strconv.FormatInt(startTime, 10)
```
chtyim: The job name should be prefixed with `"cdap-" + masterSpec.Name` to make it clear which CDAP instance the job is for. See the setTemplateValue method https://github.com/cdapio/cdap-operator/pull/9/files#diff-0c39caca86fbd994c47c05c67732106aR185

Also, the master spec name is unique without a K8s namespace, so maybe we don't need the startTimeNano to make the job name unique.

pkg/controller/cdapmaster/cdapmaster_controller.go
```diff
@@ -353,6 +395,58 @@ func logbackFromFile(t string, data interface{}) (string, error) {
...
+// Adds a component list object in state in progress. This will cause UpdateStatus to update the
+// ready type to state "false".
+func AddUpgradeComponentNotReady(master *alpha1.CDAPMaster) {
+	os := status.ObjectStatus{}
+	os.Name = "upgrade"
+	os.Status = status.StatusInProgress
+	master.Status.ComponentMeta.ComponentList.Objects = append(master.Status.ComponentMeta.ComponentList.Objects, os)
+}
+
+func LaunchJob(master *alpha1.CDAPMaster, rsrclabels map[string]string) ([]reconciler.Object, error) {
```
chtyim: BTW, it is not actually launching right? The name seems a bit misleading. Shouldn't this just merge with the `GetJobResources` below?

pkg/controller/cdapmaster/cdapmaster_controller.go
```diff
@@ -353,6 +395,58 @@ func logbackFromFile(t string, data interface{}) (string, error) {
...
+	if err != nil {
+		SetStatusFailed(master, upgradeFailedInitMessage)
+		return nil, err
+	}
+	return upgradeResources, nil
+}
+
+// Gets the job resources and sets status to not ready, since any time when we wish to add the job
+// to expected, the status is not ready.
+func GetJobResources(master *alpha1.CDAPMaster, rsrclabels map[string]string) ([]reconciler.Object, error) {
```
chtyim: Seems this is quite specific for the pre-upgrade job. I think it's better to call it `getPreUpgradeJobResources`, which I also think should be a private function instead of public

pkg/controller/cdapmaster/cdapmaster_controller.go
```diff
@@ -353,6 +395,58 @@ func logbackFromFile(t string, data interface{}) (string, error) {
 	return output.String(), nil
 }
 
+func SetStatusFailed(master *alpha1.CDAPMaster, message string) {
```
chtyim: Should be a private function?

chtyim (commented):
Some higher level comments. Still need to look at the logic in greater details.

config/crds/cdap_v1alpha1_cdapmaster.yaml
jayfjin: think some of these comments are on an outdated commit

pkg/controller/cdapmaster/cdapmaster_controller.go
jayfjin: This sets/clears additional status fields when the job is first launched

pkg/controller/cdapmaster/cdapmaster_controller.go
jayfjin: What if an upgrade is retried by the user and the job has not been removed by ttlSeconds yet? Won't the name field still be the same?

jayfjin:
Addressed comments, please take another look

pkg/controller/cdapmaster/cdapmaster_controller.go
```diff
@@ -384,9 +519,15 @@ func (b *Base) Observables(rsrc interface{}, labels map[string]string) []reconci
 	return k8s.NewObservables().
 		WithLabels(labels).
 		For(&corev1.ConfigMapList{}).
+		For(&batchv1.JobList{}).
 		Get()
 }
 
+// UpdateStatus for updating status for CDAP master
```
chtyim: It's better to move the cdapmaster_status.go file to this package and keeping the observer methods and the status update methods in separate files.

pkg/controller/cdapmaster/cdapmaster_controller.go
```diff
@@ -353,6 +394,53 @@ func logbackFromFile(t string, data interface{}) (string, error) {
...
+}
+
+// Gets the job resources and sets status to not ready, since any time when we wish to add the job
+// to expected, the status is not ready.
+func getPreUpgradeJobResources(master *alpha1.CDAPMaster, rsrclabels map[string]string) ([]reconciler.Object, error) {
+	upgradeResources, err := getPreUpgradeResources(
+		&upgradeJobSpec{
+			Image:        master.Spec.Image,
+			JobName:      getUpgradeJobName(master),
+			HostName:     getConfigName(master, "router"),
```
chtyim: This is a bit confusion why the hostname is the config name. In fact, it is a coincidence that the config map name is the same format as the router service name. The service name is actually set from the `setTemplateValue` method, using the `ServiceRouter` constant as defined in the `constants.go`. Suggest to extract the name formatting line from the `setTemplateValue` into a private method and unify it with the `getConfigName` method.

pkg/controller/cdapmaster/cdapmaster_controller.go
```diff
@@ -376,6 +464,53 @@ func (b *Base) Objects(rsrc interface{}, rsrclabels map[string]string, observed,
 		}
 	}
 
+	item := k8s.GetItem(observed, &batchv1.Job{}, getUpgradeJobName(master), "default")
+	if item == nil {
+	}
```
chtyim, line 482: What's the purpose of this empty block?

pkg/controller/cdapmaster/cdapmaster_controller.go
```diff
@@ -376,6 +464,53 @@ func (b *Base) Objects(rsrc interface{}, rsrclabels map[string]string, observed,
 		}
 	}
 
+	item := k8s.GetItem(observed, &batchv1.Job{}, getUpgradeJobName(master), "default")
```
chtyim: Why always fetch the Job? Shouldn't we check the status to see if there is upgrade in progress before getting the job?

pkg/controller/cdapmaster/cdapmaster_controller.go
```diff
@@ -353,6 +394,53 @@ func logbackFromFile(t string, data interface{}) (string, error) {
...
+// Adds a component list object in state in progress. This will cause UpdateStatus to update the
+// ready type to state "false".
+func addUpgradeComponentNotReady(master *alpha1.CDAPMaster) {
+	os := status.ObjectStatus{}
+	os.Name = "upgrade"
+	os.Status = status.StatusInProgress
+	master.Status.ComponentMeta.ComponentList.Objects = append(master.Status.ComponentMeta.ComponentList.Objects, os)
+}
+
+func launchJob(master *alpha1.CDAPMaster, rsrclabels map[string]string) ([]reconciler.Object, error) {
```
chtyim: Instead of calling it `launchJob`, better call it as `prepareNewPreUpgradeJobResources`

pkg/controller/cdapmaster/cdapmaster_controller.go
```diff
@@ -376,6 +464,53 @@ func (b *Base) Objects(rsrc interface{}, rsrclabels map[string]string, observed,
...
+		var err error
+		upgradeResources, err = getPreUpgradeJobResources(master, rsrclabels)
+		if err != nil {
+			return nil, err
+		}
+		resetImage = false
+	} else if upgradeJob != nil && upgradeJob.Status.Succeeded > 0 {
+		master.Status.ImageToUse = master.Spec.Image
+		master.Status.UserInterfaceImageToUse = master.Spec.UserInterfaceImage
+		master.Status.UpgradeJobVersion = uuid.New().String()
```
chtyim: Why generate a new uuid?

templates/upgrade-job.yaml
```diff
@@ -0,0 +1,29 @@
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: {{.Job.JobName}}
```
chtyim, line 17: Need to add the owner reference to be the CDAP CR that triggers the running of this job.

pkg/controller/cdapmaster/cdapmaster_controller.go
jayfjin: Set to empty since we check on this status before checking for the job now. Before, this could have created issues if another upgrade was made shortly after the first finished

templates/upgrade-job.yaml
jayfjin: We've set ttlSecondsAfterFinished and ttlSeconds so this job will clean up on its own. According to documentation, Jobs are expected to clean up on their own. https://kubernetes.io/docs/concepts/workloads/controllers/#job

pkg/controller/cdapmaster/cdapmaster_controller.go
jayfjin: good point, renamed getConfigName to getResourceName and used it in setTemplateValue

pkg/controller/cdapmaster/cdapmaster_controller.go
```diff
@@ -170,7 +181,7 @@ func (v *templateBaseValue) setTemplateValue(rsrc interface{}, rsrclabels map[st
 	labels.Merge(master.Labels, serviceLabels, rsrclabels)
 
 	// Set the cdap.container label. It is for service selector to route correctly
-	name := fmt.Sprintf("cdap-%s-%s", master.Name, strings.ToLower(string(serviceType)))
+	name := getResourceName(master, strings.ToLower(string(serviceType)))
```
chtyim: Since k8s requires resource name to be all lower case, it's good to move the `ToLower` inside the function.

https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names

pkg/controller/cdapmaster/cdapmaster_controller.go
```diff
@@ -370,12 +458,60 @@ func (b *Base) Objects(rsrc interface{}, rsrclabels map[string]string, observed,
...
 	for k, v := range configs {
-		resources, err = addConfigMapItem(master, getConfigName(master, k), labels, v, resources)
+		resources, err = addConfigMapItem(master, getResourceName(master, k), labels, v, resources)
 		if err != nil {
 			return nil, err
 		}
 	}
 
+	var item metav1.Object
+	if master.Status.UpgradeJobVersion == "" && master.Status.ImageToUse != master.Spec.Image {
```
chtyim: You can return early if this condition is not true as there is no upgrade needed, right? I believe it can simplify the following big if-else block.

templates/upgrade-job.yaml
chtyim: It is more than just cleanup. First, having owner reference will get the job get killed immediately on the deletion of the CDAP resource, in which resources are freed much faster. Second, it is about being a good citizen in the cluster to work nicely with other tools (e.g. a visualization tools that shows relations between resources).

pkg/controller/cdapmaster/cdapmaster_controller.go
```diff
@@ -370,12 +471,68 @@ func (b *Base) Objects(rsrc interface{}, rsrclabels map[string]string, observed,
...
+		return resources, nil
+	}
+
+	var upgradeJob *batchv1.Job
+	if item != nil {
+		upgradeJob = item.(*batchv1.Job)
+	}
+	var upgradeResources []reconciler.Object
+	resetImage := true
+	if upgradeJob == nil && !master.Status.Meta.IsConditionTrue(upgradeFailed) && imageToUseNeedsUpdate(master) {
```
chtyim: Can you add some comments about the case that each of these blocks is trying to handle?

pkg/controller/cdapmaster/cdapmaster_controller.go
```diff
@@ -370,12 +472,79 @@ func (b *Base) Objects(rsrc interface{}, rsrclabels map[string]string, observed,
...
 
+	// User has reset the CR after failure. Remove upgrade failed status.
+	if master.Spec.Image == master.Status.ImageToUse && master.Status.Meta.IsConditionTrue(upgradeFailed) {
+		master.Status.ClearCondition(upgradeFailed, upgradeResetMessage, upgradeResetMessage)
+	}
+
+	if !imageToUseNeedsUpdate(master) {
+		// Set ImageToUse if upgrade has not failed
+		if !master.Status.Meta.IsConditionTrue(upgradeFailed) {
+			master.Status.ImageToUse = master.Spec.Image
```
chtyim: This logic seems contradicting, as it is read as "if the image to use doesn't need update (line 486), update the image to use (line 489)".

pkg/controller/cdapmaster/cdapmaster_controller.go
```diff
@@ -236,9 +248,43 @@ type externalServiceValue struct {
...
+	Job *upgradeJobSpec
+}
+
+// UpgradeJobSpec defines the specification for the upgrade job
+type upgradeJobSpec struct {
+	Image        string `json:"image,omitempty"`
+	JobName      string `json:"jobName,omitempty"`
+	HostName     string `json:"hostName,omitempty"`
+	BackoffLimit int32  `json:"backoffLimit,omitempty"`
+	ReferentName string `json:referentName,omitempty"`
```
chtyim: Seems like you missed a `"`. Should be `json:"referentName,omitempty"`

pkg/controller/cdapmaster/cdapmaster_controller.go
```diff
@@ -370,12 +472,79 @@ func (b *Base) Objects(rsrc interface{}, rsrclabels map[string]string, observed,
...
+		// Pre upgrade job has already been initialized, but has not finished
+		var err error
+		upgradeResources, err = getPreUpgradeJobResources(master, rsrclabels)
+		if err != nil{
+			return nil, err
+		}
+		addUpgradeComponentNotReady(master)
+	}
+
+	if upgradeResources != nil {
```
chtyim: You can save this if by moving the `resources = append(resources, upgradeResources...)` inside the last `else` block.

pkg/controller/cdapmaster/cdapmaster_controller.go
```diff
@@ -370,10 +472,74 @@ func (b *Base) Objects(rsrc interface{}, rsrclabels map[string]string, observed,
...
+			return nil, err
+		}
+	}
+
+	// User has reset the CR after failure. Remove upgrade failed status.
+	if master.Spec.Image == master.Status.ImageToUse && master.Status.Meta.IsConditionTrue(upgradeFailed) {
+		master.Status.ClearCondition(upgradeFailed, upgradeResetMessage, upgradeResetMessage)
+	}
+
+	if !isImageUpdatePending(master) {
```
chtyim: Since there is only one usage of this method, I think it's better inline the logic here directly (with comment), which make it clearer for the condition and intention is.

pkg/controller/cdapmaster/cdapmaster_controller.go
```diff
@@ -370,10 +466,77 @@ func (b *Base) Objects(rsrc interface{}, rsrclabels map[string]string, observed,
...
+			master.Status.ClearCondition(upgradeFailed, upgradeResetMessage, upgradeResetMessage)
+		}
+		return resources, nil
+	}
+
+	// Early return if image does not need updating. Update imageToUse since we are in a non
+	// failure case.
+	if master.Status.ImageToUse == "" || master.Status.ImageToUse == master.Spec.Image {
+		master.Status.ImageToUse = master.Spec.Image
+		master.Status.UserInterfaceImageToUse = master.Spec.UserInterfaceImage
```
chtyim, line 489: The intention of having two docker images for the backend and the UI is to allow them upgrade separately. It may not be an issue now by assuming they get upgraded together, but it's good to note that it may be different in future.

pkg/controller/cdapmaster/cdapmaster_controller.go
```diff
@@ -370,10 +466,77 @@ func (b *Base) Objects(rsrc interface{}, rsrclabels map[string]string, observed,
...
+		return resources, nil
+	}
+
+	var upgradeResources []reconciler.Object
+	if master.Status.UpgradeJobVersion == "" {
```
chtyim: You can flip the condition to reduce code duplication in this if block and the `upgradeJob == nil` case.

```java
var upgradeJob *batchv1.Job
if master.Status.UpgradeJobVersion != "" {
	item := k8s.GetItem(observed, &batchv1.Job{}, getUpgradeJobName(master), "default")
	if item != nil {
		upgradeJob = item.(*batchv1.Job)
	}
}
if upgradeJob == nil {
		// Add a new upgrade job, and set ready status to false
		var err error
		upgradeResources, err = prepareNewPreUpgradeJobResources(master, rsrclabels)
		if err != nil {
			return nil, err
		}
		addUpgradeComponentNotReady(master)
		return append(resources, upgradeResources...), nil
}
// Otherwise upgradeJob exists, handle succeed / failure case.
```

chtyim (approved):
The logic looks good. Just one comment about reducing code duplication. Please squash the commits before merging.

pkg/controller/cdapmaster/cdapmaster_controller.go
jayfjin: oh nice!

PR #12, opened 2019-08-26
Add logic for post upgrade

jayfjin:
- Add logic to start a post upgrade job after the operator is back up in a ready state
- Remove UpgradeJobVersion, since the UUID is too long to include in the job name, and startTimeMillis effectively fills the same purpose.
- Add logic to handle downgrade case - skip pre/post upgrade jobs

### DIFF
--- a/pkg/controller/cdapmaster/cdapmaster_controller.go
+++ b/pkg/controller/cdapmaster/cdapmaster_controller.go
@@ -473,39 +473,42 @@ func getPostUpgradeJobResources(master *alpha1.CDAPMaster, rsrclabels map[string
 	return getUpgradeJobResources(master, rsrclabels, master.Status.UpgradeStartTimeMillis, getPostUpgradeJobName(master))
 }
 
-func getVersion(a string) string {
+func getVersion(a string) *version {
 	av := strings.Split(a, ":")
-	if len(av) == 1 {
-		 return latestVersion
+	if len(av) == 1 || av[1] == latestVersion {
+		 return &version{isLatest: true}
 	}
-	return av[1]
+	return &version{isLatest:false, versionList: strings.Split(av[1], ".")}
 }
 
-func compareVersion(a, b string) int {
-	versiona := getVersion(a)
-	versionb := getVersion(b)
+// Represents an image version
+type version struct {
+	// Boolean if the version is "latest"
+	isLatest bool
+	// List of version integers, starting from major
+	versionList []string
+}
 
-	if latestVersion == versiona && latestVersion == versionb {
+func compareVersion(versiona, versionb *version) int {
+	if versiona.isLatest && versionb.isLatest {
 		return 0
-	} else if latestVersion == versiona {
+	} else if versiona.isLatest{
 		return -1
-	} else if latestVersion == versionb {
+	} else if versionb.isLatest {
 		return 1
 	}
 
-	as := strings.Split(versiona, ".")
-	bs := strings.Split(versionb, ".")
-	loopMax := len(bs)
-	if len(as) > len(bs) {
-		loopMax = len(as)
+	loopMax := len(versionb.versionList)
+	if len(versiona.versionList) > len(versionb.versionList) {
+		loopMax = len(versiona.versionList)
 	}
 	for i := 0; i < loopMax; i++ {
 		var x, y string
-		if len(as) > i {
-			x = as[i]
+		if len(versiona.versionList) > i {
+			x = versiona.versionList[i]
 		}
-		if len(bs) > i {
-			y = bs[i]
+		if len(versionb.versionList) > i {
+			y = versionb.versionList[i]
 		}
 		xi,_ := strconv.Atoi(x)
 		yi,_ := strconv.Atoi(y)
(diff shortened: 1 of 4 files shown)


pkg/controller/cdapmaster/cdapmaster_controller.go
```diff
@@ -458,25 +462,90 @@ func prepareNewPreUpgradeJobResources(master *alpha1.CDAPMaster, rsrclabels map[
...
+	return getUpgradeJobResources(master, rsrclabels, 0, getUpgradeJobName(master))
+}
+
+// Gets post upgrade job resources
+func getPostUpgradeJobResources(master *alpha1.CDAPMaster, rsrclabels map[string]string) ([]reconciler.Object, error) {
+	return getUpgradeJobResources(master, rsrclabels, master.Status.UpgradeStartTimeMillis, getPostUpgradeJobName(master))
+}
+
+func compareVersion(a, b string) int {
+	av:= strings.Split(a, ":")
```
chtyim: space before `:`.

pkg/controller/cdapmaster/cdapmaster_controller.go
```diff
@@ -458,25 +462,90 @@ func prepareNewPreUpgradeJobResources(master *alpha1.CDAPMaster, rsrclabels map[
...
+	bv := strings.Split(b, ":")
+
+	var versiona, versionb string
+	if len(av) == 1 {
+		versiona = latestVersion
+	} else {
+		versiona = av[1]
+	}
+	if len(bv) == 1 {
```
chtyim: refactor repetitive code into function: `leftVersion := getVersion(leftImage)`

pkg/controller/cdapmaster/cdapmaster_controller.go
```diff
@@ -458,25 +462,90 @@ func prepareNewPreUpgradeJobResources(master *alpha1.CDAPMaster, rsrclabels map[
...
+
+	if latestVersion == versiona && latestVersion == versionb {
+		return 0
+	} else if latestVersion == versiona {
+		return -1
+	} else if latestVersion == versionb {
+		return 1
+	}
+
+	as := strings.Split(versiona, ".")
```
chtyim: Better encapsulate the parsing logic into a function and have it return a `version struct`

pkg/controller/cdapmaster/cdapmaster_controller.go
```diff
@@ -515,7 +584,45 @@ func (b *Base) Objects(rsrc interface{}, rsrclabels map[string]string, observed,
...
+		master.Status.ImageToUse = master.Spec.Image
+		master.Status.UserInterfaceImageToUse = master.Spec.UserInterfaceImage
+		master.Status.SetCondition(postUpgradeFinished, upgradeJobSkippedMessage, upgradeJobSkippedMessage)
+		return resources, nil
+	}
+
+	// Post upgrade case, in which the image has updated successfully and all pods are back up
+	if master.Status.UpgradeStartTimeMillis != 0 && master.Status.IsReady() {
+		var postUpgradeJob *batchv1.Job
+		item := k8s.GetItem(observed, &batchv1.Job{}, getPostUpgradeJobName(master), "default")
```
chtyim: I didn't notice that we've hardcoded to "default" namespace. We shouldn't be and should be running the job in the same namespace as CDAP does (not the controller namespace).

pkg/controller/cdapmaster/cdapmaster_controller.go
jayfjin: The version info is just a list of ints currently, so would prefer to keep it as is for simplicity

chtyim: The problem is that it makes the following `for` loop very cryptic to look at. It is not programming by intend.

pkg/controller/cdapmaster/cdapmaster_controller.go
```diff
@@ -458,25 +463,90 @@ func prepareNewPreUpgradeJobResources(master *alpha1.CDAPMaster, rsrclabels map[
...
+	}
+	for i := 0; i < loopMax; i++ {
+		var x, y string
+		if len(versiona.versionList) > i {
+			x = versiona.versionList[i]
+		}
+		if len(versionb.versionList) > i {
+			y = versionb.versionList[i]
+		}
+		xi,_ := strconv.Atoi(x)
```
chtyim, line 514: What if the version is not a number? Say v6.1.0

jayfjin: We need to place restrictions on what the image/version can look like, otherwise it would be impossible to parse. Since only "latest" or "X.X.X.X" are used currently, I think this is ok

chtyim: Or we can just split by ".", then for each part of the split, if it can be parsed as number, then we treat it as number, otherwise leave it as string. We can make the assumption that number is always greater than string, and if both are strings, then just do string comparision.

jayfjin: This is almost the behavior of the current code, since strconv returns 0 on failure along with an err we don't save. The only difference is if both are strings, we return equals here which I think is fine... since wanting to string compare parts of a version is not actually a use case

chtyim: Sounds good. I didn't notice the Atoi will return `0` on such case.

chtyim (approved):

PR #13, opened 2019-08-28
(CDAP-15802) Add RuntimeClassName support

chtyim:


jayfjin (commented):
Does upgrade job need this as well?

chtyim:
I don't think the upgrade job need this for now. The support for runtime class is mainly for sandboxing workload, in which the upgrade job is not running user code in CDAP, hence there is no sandboxing need.

jayfjin (approved):

PR #23, opened 2020-01-31
Restructure operator to allow easy support for  multi-container pod deployment 

wyzhang:
Re-structuring operator to introduce "internal" crd (sort of), so CR gets mapped to "internal" CR which  then used to fill the template and generate stateful/deployment/services. 

spec.go
- This is where "internal" CRD is defined 

Note that 
- Manual tested. UI loads. 
- Upgrade/downgrade logic is missing. Yet to be added.
- Some fields in template haven't been set or set correctly yet (e.g. env). Will fix in follow-up changes.


templates/cdap-ui-deployment.yaml
```diff
@@ -14,55 +14,52 @@
...
-  namespace: {{.Master.Namespace}}
-{{if .Labels}}
+  name: {{.Base.Name}}
+  namespace: {{.Base.Namespace}}
+{{if .Base.Labels}}
   labels:
-    {{range $k,$v := .Labels}}
-    {{$k}}: '{{$v}}'
+    {{range $k,$v := .Base.Labels}}
+    {{$k}}: {{$v}}
```
barney-s: dont you need the single quotes ?

templates/cdap-ui-deployment.yaml
```diff
@@ -14,55 +14,52 @@
...
-{{else}}
-  replicas: 1
-{{end}}
+  replicas: {{.Base.Replicas}}
   template:
     metadata:
       labels:
-        {{range $k,$v := .Labels }}
-        {{$k}}: '{{$v}}'
+        {{range $k,$v := .Base.Labels }}
```
barney-s: same single quote comment

templates/cdap-ui-deployment.yaml
```diff
@@ -14,55 +14,52 @@
...
-    {{if .Service.PriorityClassName}}
+    {{if .Base.PriorityClassName}}
       priorityClassName: {{.Service.PriorityClassName}}
     {{end}}
+    {{range $c := .Containers}}
       containers:
-        - image: {{.Master.Status.UserInterfaceImageToUse}}
+        - name: {{$c.Name}}
+          image: {{$c.Image}}
           workingDir: "/opt/cdap/ui"
```
barney-s: is working dir same for all containers (non ui ones)

barney-s (commented):
first pass

templates/cdap-ui-deployment.yaml
wyzhang: Yes. For this specific template, it is used by our UI service, which will be a normal single-container pod.

templates/cdap-ui-deployment.yaml
wyzhang: Not sure. Some of our existing template use single quotes. Others don't. Seem to work without them, thus removing to make it consistent.

From quick search
https://stackoverflow.com/questions/19109912/do-i-need-quotes-for-strings-in-yaml
looks to be quotes are in generally not required

wyzhang:
Made a bunch of major revisions. The change getting much bigger. 
Major changes are: 
* Re-wrote the upgrade/downgrade logic 
* Added unit tests. Decent test coverage. More to be added. 
* Broke code into multiple files

With all these changes, it achieves the same functionality as the current "develop" branch + configurable # of pods deployment

wyzhang:
One more major changes forgot to mention:
- Removed the UI yaml template. Made the deployment template flexible that can be adopted by UI

wyzhang:
Some code refactoring: using reflect to get pointer to struct field, thus simplify the code

controllers/version_update.go
```diff
@@ -0,0 +1,489 @@
...
+	}
+	if len(curVersion.rawString) == 0 {
+		setImageToUse(master)
+		return []reconciler.Object{}, nil
+	}
+
+	switch compareVersion(curVersion, newVersion) {
+	case -1:
+		// Upgrade:
+		updateStatus.clearAllConditions(master)
```
jayfjin, line 70: Is this correct? Won't this clear the status each reconcile loop, thus defeating the purpose of tracking the state.

wyzhang: Updated the code based on offline discussion. 

At the high level, make cdap operator not retrying upgrade. So if upgrade fails, leave all conditions in their state for user to exam. 

So the logic to check the upgrade status and react to it is:
if (upgrade inflight) { wait }
if (upgrade failed condition is set) {
  if (spec.image == status.image) {
     return upgrade_failed
  } else {
    return bad_state (because image upgraded but post-upgrade processing failed)
  }
}
return upgrade_succeeded.

jayfjin: lgtm

jayfjin (approved):
lgtm for upgrade based on our discussions

PR #25, opened 2020-02-21
Fixed docker build file.

pandyajay10:
Missing vendor directory copy step. Verified locally.

Previous state: Did not build successfully.
Current State: Builds successfully.


Dockerfile
```diff
@@ -5,6 +5,8 @@ WORKDIR /workspace
 # Copy the Go Modules manifests
 COPY go.mod go.mod
 COPY go.sum go.sum
+COPY vendor/ vendor/
+
```
wyzhang, line 9: As discussed offline, we should copy template as well. Or just copy everything.

PR #28, opened 2020-03-10
Quote many user provided values in Yaml templates

### LINKED ISSUE #27: Labels with only numbers as value breaks deployment/sts templates
I have a label automatically added to the cdap resource with a number as its value, it looks like:

```
kapp.k14s.io/app=1583512484380481917
```

This breaks the `cdap-deployment.yaml` and `cdap-sts.yaml` templates, because label values aren't quoted, and may be interpreted as a number (instead of a string). I'd image the same would happen for truthy/falsy values.

Here's an example error message from the operator:

```
E0310 17:46:06.200716       1 genericreconciler.go:51] Failed: [*v1alpha1.CDAPMaster/default/cdap] gathering expected resources. templates/cdap-sts.yaml:v1.StatefulSet.ObjectMeta: v1.ObjectMeta.Labels: ReadStr
ing: expects " or n, but found 1, error found in #10 byte of ...|.io/app":15835124843|..., bigger context ...|urce-namespace":"default","kapp.k14s.io/app":1583512484380481917,"kapp.k14s.io/association":"v1.|..
.
2020/03/10 17:46:06 *v1alpha1.CDAPMaster/default/cdap(cmpnt:*cdapmaster.Messaging)  } reconciling component
E0310 17:46:06.200746       1 genericreconciler.go:114] error reconciling *v1alpha1.CDAPMaster/default/cdap. templates/cdap-sts.yaml:v1.StatefulSet.ObjectMeta: v1.ObjectMeta.Labels: ReadString: expects " or n,
 but found 1, error found in #10 byte of ...|.io/app":15835124843|..., bigger context ...|urce-namespace":"default","kapp.k14s.io/app":1583512484380481917,"kapp.k14s.io/association":"v1.|...
```

I think most (all?) values in these templates should be quoted, to avoid these types of issues. At least the label values should be.

Happy to put in a PR for this, but wanted feedback first. I'd like to go through and quote everything that has foreign data in it.

jess-sol:
This fixes issue #27, where K8S would implicitly assume a value was a number/boolean/etc. instead of a string in the given Yaml template, due to the value not being quoted.

### DIFF
--- a/templates/cdap-deployment.yaml
+++ b/templates/cdap-deployment.yaml
@@ -19,21 +19,21 @@ metadata:
   {{if .Base.Labels}}
   labels:
     {{range $k,$v := .Base.Labels}}
-    {{$k}}: {{$v}}
+    {{$k}}: "{{$v}}"
     {{end}}
   {{end}}
 spec:
   selector:
     matchLabels:
       {{range $k,$v := .Base.Labels}}
-      {{$k}}: {{$v}}
+      {{$k}}: "{{$v}}"
       {{end}}
   replicas: {{.Base.Replicas}}
   template:
     metadata:
       labels:
         {{range $k,$v := .Base.Labels }}
-        {{$k}}: {{$v}}
+        {{$k}}: "{{$v}}"
         {{end}}
     spec:
       {{if .Base.ServiceAccountName}}
@@ -41,7 +41,7 @@ spec:
       {{end}}
       nodeSelector:
         {{range $k,$v := .Base.NodeSelector}}
-        {{$k}}: {{$v}}
+        {{$k}}: "{{$v}}"
         {{end}}
       {{if .Base.RuntimeClassName}}
       runtimeClassName: {{.Base.RuntimeClassName}}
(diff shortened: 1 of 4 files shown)


jess-sol:
@googlebot I signed it!

templates/cdap-deployment.yaml
```diff
@@ -19,29 +19,29 @@ metadata:
   {{if .Base.Labels}}
   labels:
     {{range $k,$v := .Base.Labels}}
-    {{$k}}: {{$v}}
+    "{{$k}}": "{{$v}}"
```
chtyim: YAML keys are always string. Quote is not needed.

jess-sol: I'm not sure that's the case, from the Yaml spec:
> The content of a mapping node is an unordered set of key: value node pairs, with the restriction that each of the keys is unique. YAML places no further restrictions on the nodes. In particular, keys may be arbitrary nodes, the same node may be used as the value of several key: value pairs, and a mapping could even contain itself as a key or a value (directly or indirectly).

[Source](https://yaml.org/spec/1.2/spec.html#id2764044)

That said, I'm happy to remove the quotes, I don't really think they're necessary for keys, it was just an overabundance of caution.

chtyim (approved):

PR #29, opened 2020-03-17
Remove numPods option from CRD, since the use case is no longer there.

wyzhang:
Remove "numPods" option from CRD, since the use case is no longer there. 
But keep the underlying implication there so we can use it to colocate services in the same pod if needed in the future.


controllers/deployment.go
```diff
@@ -121,11 +77,9 @@ func buildDeploymentPlanSpec(master *v1alpha1.CDAPMaster, labels map[string]stri
 		return &DeploymentPlanSpec{}, nil
 	}
 
-	// Get deployment plan depending on the number of pods desired
+	// Get the deployment plan. At the moment, it only supports numPods = 0 (i.e. one service per pod).
```
chtyim, line 80: Why zero means one? Should it better assign it to `1`?

chtyim (approved):
Just one question

controllers/deployment.go
wyzhang: numPods here refers to the total number of pods used for deploying CDAP. 
numPods == 1--> all services in 1 pod 
numPods == 2--> UI service in 1 pod and backend services in 1 pod
...

numPods == 8 -> basically means each service run in its own pod, but instead of having user to cont the total number of services and set it to 8 to achieve one service per pod. Setting numPods = 0 or omitting it indicates one service per pod.

PR #32, opened 2020-03-21
Support operator upgrade backward compatibility.

wyzhang:
Motivation:
Without this change, upgrading operator from old version to the
new version in a kubernetes cluster with CDAP already deployed
would cause new operator unable to identify already deployed
services. This is because the k8s objects created are tagged
with label <package_name>.<handler_struct_name> which are different
in old and new version.

Fix:
Adding back reconciling handlers from tags/v1.0 to proper clean
up k8s objects created by old version of the operator. The
new operator will re-deploy CDAP services.

### DIFF
--- a/controllers/cdapmaster/cdapmaster_controller.go
+++ b/controllers/cdapmaster/cdapmaster_controller.go
@@ -1,0 +1,187 @@
+/*
+Copyright 2020 The CDAP Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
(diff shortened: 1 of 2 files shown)


chtyim:
Do we support downgrade too?

wyzhang:
Not at the moment, but we can add similar logic to tags/v1.0.

Do we need to support downgrade? I am not sure if I see a strong use case for that. If the new operator can handle old CR properly (which we have tested and verified), there seems to be no strong reason of downgrading operator. 

If we do feel strongly to support downgrade, will open a separate PR for that.

chtyim (approved):

pandyajay10 (approved):

PR #39, opened 2020-04-10
Fix missing quotes in json tag for newly added ConfigMapVolumes

wyzhang:


wyzhang:
Also fixing a bug in template that sets the wrong name for the volume. 

Tested manually end-to-end

pandyajay10 (approved):

PR #51, opened 2020-07-24
increase activeDeadlineSeconds setting for pre/post upgrade job to 6h and increasing retries to 20

pandyajay10:
Increased activeDeadlineSeconds parameter for upgrade job configuration to 6h to have an upper bound on job runtime but it should not be possible for upgrade job to reach this time. Something seriously would have gone wrong if upgrade job retries are not exhausted before 6h deadline.

Also increased number of retry counts for pre and post upgrade job to 20 as for a CDAP cluster with many jobs running/streaming jobs running, it might take a long time to stop them.

### DIFF
--- a/templates/upgrade-job.yaml
+++ b/templates/upgrade-job.yaml
@@ -73,5 +73,8 @@ spec:
       {{end}}
       restartPolicy: Never
   backoffLimit: {{.Job.BackoffLimit}}
-  activeDeadlineSeconds: 600
+  # Set job deadline to 6h as pre/post upgrade job can take a lot of time to complete in scenarios like instance with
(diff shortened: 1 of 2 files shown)


templates/upgrade-job.yaml
```diff
@@ -73,5 +73,5 @@ spec:
       {{end}}
       restartPolicy: Never
   backoffLimit: {{.Job.BackoffLimit}}
-  activeDeadlineSeconds: 600
+  activeDeadlineSeconds: 21600
```
wyzhang, line 79: Add comments to explain the rational behind bumping it up

wyzhang (approved):

PR #55, opened 2020-11-14
allow http port in image name

kwop:
Adding an http port to cdap container image failed, due to array spliting limitation.

Ex : gcr.io:1337/cdapio/cdap:1.3.3.7

### DIFF
--- a/controllers/version_update.go
+++ b/controllers/version_update.go
@@ -372,18 +372,19 @@ func parseImageString(imageString string) (*Version, error) {
 		return &Version{}, nil
 	}
 	splits := strings.Split(imageString, ":")
-	if len(splits) != 2 {
-		return nil, fmt.Errorf("failed to parse image string %s, not in expected format of xxx:xxx", imageString)
+	if len(splits) > 3 {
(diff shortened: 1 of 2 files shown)


kwop:
@googlebot I signed it!

controllers/version_update.go
```diff
@@ -372,18 +372,19 @@ func parseImageString(imageString string) (*Version, error) {
 		return &Version{}, nil
 	}
 	splits := strings.Split(imageString, ":")
-	if len(splits) != 2 {
-		return nil, fmt.Errorf("failed to parse image string %s, not in expected format of xxx:xxx", imageString)
+	if len(splits) > 3 {
```
chtyim, line 375: It is better to change to use `LastIndex` to look for the last `:` to extract the version

controllers/version_update_test.go
```diff
@@ -17,13 +17,28 @@ import (
 
 var _ = Describe("Controller Suite", func() {
 	Describe("Parsing image string", func() {
-		It("Parse empty image string", func() {
+		It("Parse empty image string with http port", func() {
```
wyzhang, line 20: Any reason to change the description of this test?

PR #56, opened 2020-11-14
allow optionnal imagePullSecrets on deployments

kwop:
Allow to fetch images from private repositories.


controllers/spec.go
```diff
@@ -245,10 +247,11 @@ func newStorageSpec(storageClassName string, storageSize string) *StorageSpec {
 
 // For StatefulSet
 type StatefulSpec struct {
-	Base           *BaseSpec        `json:"Base,inline"`
-	InitContainers []*ContainerSpec `json:"initContainer,omitempty"`
-	Containers     []*ContainerSpec `json:"containers,omitempty"`
-	Storage        *StorageSpec     `json:"storage,omitempty"`
+	Base             *BaseSpec                     `json:"Base,inline"`
+	ImagePullSecrets []corev1.LocalObjectReference `json:"imagePullSecrets,omitempty"`
```
chtyim, line 251: Shoudln't need to add this if already added in the BaseSpec

wyzhang: +1

PR #59, opened 2020-12-20
[CDAP-17466] Authentication functionality for CDAP on Kubernetes setup

edvinas-maciulis:
Design document: https://wiki.cask.co/display/CE/Authentication+feature+for+CDAP+on+Kubernetes+deployment
JIRA Ticket: https://cdap.atlassian.net/browse/CDAP-17466

PR #61, opened 2021-02-26
[CDAP-17466] Added supported for optionally running an Authentication service in k8s.

chtyim:
This one replaces #59 with test fixes.


api/v1alpha1/cdapmaster_types.go
```diff
@@ -146,7 +163,7 @@ type MessagingSpec struct {
 
 // MetadataSpec defines the specification for the Metadata service
 type MetadataSpec struct {
-	CDAPStatefulServiceSpec `json:",inline"`
+	CDAPScalableServiceSpec `json:",inline"`
```
wyzhang, line 166: Thanks for fixing the mistake here.

api/v1alpha1/cdapmaster_types.go
```diff
@@ -98,6 +107,14 @@ type CDAPServiceSpec struct {
...
 	// Env is a list of environment variables for the master service container.
 	Env []corev1.EnvVar `json:"env,omitempty"`
+	// ConfigMapVolumes defines a map from ConfigMap names to volume mount path for this service
+	// Key is the configmap object name. Value is the mount path.
+	// This adds ConfigMap data to the directory specified by the volume mount path.
+	ConfigMapVolumes map[string]string `json:"configMapVolumes,omitempty"`
+	// SecretVolumes defines a map from Secret names to volume mount path for this service
+	// Key is the secret object name. Value is the mount path.
+	// This adds Secret data to the directory specified by the volume mount path.
+	SecretVolumes map[string]string `json:"secretVolumes,omitempty"`
```
wyzhang, line 117: Are these used for overriding those settings in CDAPMasterSpec ? If so, mind updating the comment to state that explicitly?

If I recall correctly, ConfigMapVolumes defined in CDAPMasterSpec will mount config map to all statefulsets and deployments.

controllers/deployment.go
```diff
@@ -94,6 +95,7 @@ func buildDeploymentPlanSpec(master *v1alpha1.CDAPMaster, labels map[string]stri
 	spec := newDeploymentPlanSpec()
 	// Build statefulsets
 	for k, v := range serviceGroups.stateful {
+
```
wyzhang: Remove the redundant new line?

templates/cdap-deployment.yaml
```diff
@@ -106,6 +106,10 @@ spec:
             - name: cdap-cm-vol-{{$k}}
               mountPath: {{$v}}
             {{end}}
+            {{range $k,$v := $.Base.SecretVolumes}}
+            - name: cdap-secret-vol-{{$k}}
```
wyzhang: Consider shortening it to cdap-se-vol-

Recall I ran into name string length limit issue when calling it cdap-configmap-vol-

templates/cdap-sts.yaml
```diff
@@ -146,6 +146,10 @@ spec:
             - name: cdap-cm-vol-{{$k}}
               mountPath: {{$v}}
             {{end}}
+            {{range $k,$v := $.Base.SecretVolumes}}
+            - name: cdap-secret-vol-{{$k}}
```
wyzhang: Same here.

controllers/spec.go
```diff
@@ -148,7 +150,16 @@ func newBaseSpec(master *v1alpha1.CDAPMaster, name string, labels map[string]str
 	s.CConf = cconf
 	s.HConf = hconf
 	s.SysAppConf = sysappconf
-	s.ConfigMapVolumes = master.Spec.ConfigMapVolumes
+
+	s.ConfigMapVolumes = make(map[string]string)
+	for k, v := range master.Spec.ConfigMapVolumes {
+		s.ConfigMapVolumes[k] = v
+	}
```
wyzhang: Agree that better to deep copy maps. Curious if shadow copy caused any sigsegv during your test, thus making this change?

wyzhang: Mind helping with converting the label copy to deep copy as well while you are here. Or feel free to leave as is that I can clean it up across the entire codebase in a separate change

wyzhang (approved):

api/v1alpha1/cdapmaster_types.go
chtyim: It is not for overriding, but appending/merging.

controllers/spec.go
chtyim: Done.

I haven't tried shallow copy, but I think it is better to keep the `master` spec not mutated.

templates/cdap-deployment.yaml
chtyim: The name limit is 253 characters. Did we run issue before?

controllers/deployment.go
chtyim: done

PR #66, opened 2021-05-27
[CDAP-18024] Add Support for Running as Non-Root User in Kubernetes

dli357:
This feature leverages Kubernetes's [SecurityContext](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/) feature in order to set a user ID, group ID, and filesystem access group ID for all the pods managed by CDAP operator.

The user ID and group ID define what user/group the pod will run with, and the filesystem access group ID defines which groups the volumes mounted onto the pod are accessible for. The filesystem access group ID is automatically applied to the user running in the pod.

For CDAP to function properly, the user assigned to ID 1000 must also be defined a valid user ID and group ID using the `useradd` and `groupadd` commands.

This is dependent on the following CDAP PR: https://github.com/cdapio/cdap/pull/13432


controllers/constants.go
```diff
@@ -49,7 +49,7 @@ const (
 	// cconf and hconf
 	confExploreEnabled        = "explore.enabled"
 	confLocalDataDirKey       = "local.data.dir"
-	confLocalDataDirVal       = "/data"
+	confLocalDataDirVal       = "/home/cdap/data"
```
wyzhang: Adding a comment saying that operator assumes that a user "cdap" exists in the image for this to worker properly.

controllers/constants.go
```diff
@@ -64,6 +64,11 @@ const (
 	labelInstanceKey        = "cdap.instance"
 	labelContainerKeyPrefix = "cdap.container."
 
+	// kubernetes security context
+	defaultSecurityContextUID   = 1000
+	defaultSecurityContextGID   = 1000
+	defaultSecurityContextFSGID = 2000
```
wyzhang, line 70: As discussed offline, would be better to make security settings configurable via cdap-cr. When they are unset, default to current behavior of running as root.

controllers/constants.go
dli357: Fixed by removing the hardcoded value.

controllers/constants.go
dli357: Fixed by making it configurable.

controllers/cdapmaster_controller.go
```diff
@@ -122,8 +122,6 @@ func ApplyDefaults(resource interface{}) {
 	if spec.Config == nil {
 		spec.Config = make(map[string]string)
 	}
-	// Set the local data directory
-	spec.Config[confLocalDataDirKey] = confLocalDataDirVal
```
wyzhang, line 126: set local data dir if absent in the cdap-cr's config section

templates/cdap-deployment.yaml
```diff
@@ -110,6 +117,12 @@ spec:
             - name: cdap-se-vol-{{$k}}
               mountPath: {{$v}}
             {{end}}
+          {{if $.Base.SecurityContext}}
+          securityContext:
+            privileged: {{$.Base.SecurityContext.Privileged}}
```
wyzhang: If some fields of security are left unspecified, I think we will use default value for them. Just wanted to make sure we thought about it. Default values look fine to me. Please double check

wyzhang (approved):

controllers/cdapmaster_controller.go
dli357: Done

templates/cdap-deployment.yaml
dli357: Good point, we need to be careful of the default behaviors. I've gone ahead and modified it so that:
 - If the integer values (runAsUser, runAsGroup, and fsGroup) are not specified (or zero) in the cdap-cr, no value is specified in the resulting yaml.
 - If the boolean values (allowPrivilegeEscalation, privileged, runAsNonRoot, readOnlyFilesystem) are not specified in the cdap-cr, they are explicitly set to default values defined by the Kubernetes API, i.e. allowPrivilegeEscalation = true, privileged = false, runAsNonRoot = false, readOnlyFilesystem = false.

PR #70, opened 2021-06-17
[CDAP-18079 CDAP-18004] Add Support for Worker Secrets

dli357:
Sets CConfiguration options for mounting secrets in worker pods including preview runner pods and task worker pods. Dependent on https://github.com/cdapio/cdap/pull/13485

### DIFF
--- a/controllers/cdapmaster_controller.go
+++ b/controllers/cdapmaster_controller.go
@@ -133,9 +133,20 @@ func ApplyDefaults(resource interface{}) {
 		spec.Config[confLocalDataDirKey] = confLocalDataDirVal
 	}
 
-	// Set twill.security.secret.disk.name to be consistent with securitySecret if not overwritten.
-	if _, ok := spec.Config[confTwillSecuritySecretDiskName]; !ok {
-		spec.Config[confTwillSecuritySecretDiskName] = spec.SecuritySecret
+	// Set security secret disk names to be consistent with securitySecret if not overwritten.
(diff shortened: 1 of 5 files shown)


controllers/cdapmaster_controller.go
```diff
@@ -133,9 +133,20 @@ func ApplyDefaults(resource interface{}) {
 		spec.Config[confLocalDataDirKey] = confLocalDataDirVal
 	}
 
-	// Set twill.security.secret.disk.name to be consistent with securitySecret if not overwritten.
-	if _, ok := spec.Config[confTwillSecuritySecretDiskName]; !ok {
-		spec.Config[confTwillSecuritySecretDiskName] = spec.SecuritySecret
+	// Set security secret disk names to be consistent with securitySecret if not overwritten.
+	if _, ok := spec.Config[confTwillSecurityMasterSecretDiskName]; !ok {
```
wyzhang: nit: just to avoid any confusion, wrap this within

if len(spec.SecuritySecret) > 0 {
  // override 
}

wyzhang (approved):

controllers/cdapmaster_controller.go
dli357: We only want to override if the cdap-cr doesn't already have that config set right?

wyzhang: Yes. 

But if security secret is empty in cdap-cr, we don't need to set these configures to empty again, right?

dli357: Fixed, thanks!

PR #73, opened 2021-10-20
[CDAP-18568] Support CustomVolumes and CustomVolumeMounts for CDAP System Services

dli357:
This feature adds the ability for a user to directly specify custom k8s volumes and volume mounts in the CDAP CR.

Custom volumes and volume mounts specified in the upper-level CR will be applied to all services, and custom volumes and volume mounts specified in the particular service will only be applied to that service. Custom volume mounts for a service are applied to all containers and init containers for that service.


dvitiiuk:
Checked that this change satisfying the next requirements (which caused the creation of #72):
 - managing of file-based secure store on a separated volume
 - storing RBAC-related artifacts on a separated volume.

api/v1alpha1/cdapmaster_types.go
```diff
@@ -93,6 +93,10 @@ type CDAPMasterSpec struct {
 	Authentication *AuthenticationSpec `json:"authentication,omitempty"`
 	// SecurityContext defines the security context for all pods for all services.
 	SecurityContext *SecurityContext `json:"securityContext,omitempty"`
+	// CustomVolumes defines a list of custom volumes for the service.
+	CustomVolumes []corev1.Volume `json:"customVolumes,omitempty"`
```
wyzhang: Does this require yaml to specify an empty array like: 

CustomerVolumes: []

Or it is fine to omit CustomerVolumes in the yaml ?

controllers/deployment.go
```diff
@@ -326,6 +338,28 @@ func buildStatefulSetsObject(spec *StatefulSpec) (*reconciler.Object, error) {
...
+		return nil, fmt.Errorf("failed to convert object to k8s object")
+	}
+	statefulSetObj, ok := k8sObj.Obj.(*appsv1.StatefulSet)
+	if !ok {
+		return nil, fmt.Errorf("failed to convert meta object to statefulset object")
+	}
+	if err := addVolumeToPodSpec(&statefulSetObj.Spec.Template.Spec, spec.Base.CustomVolumes); err != nil {
+		return nil, err
+	}
+	for ind, _ := range statefulSetObj.Spec.Template.Spec.InitContainers {
```
wyzhang: nit: ind => either idx or i. 

Have never seen "ind" as abbreviation for index

controllers/deployment.go
```diff
@@ -326,6 +338,28 @@ func buildStatefulSetsObject(spec *StatefulSpec) (*reconciler.Object, error) {
...
+		if err := addVolumeMountToContainer(&statefulSetObj.Spec.Template.Spec.InitContainers[ind], spec.Base.CustomVolumeMounts); err != nil {
+			return nil, err
+		}
+	}
+	for ind, _ := range statefulSetObj.Spec.Template.Spec.Containers {
```
wyzhang: nit: same here ind=> idx

controllers/deployment.go
```diff
@@ -335,9 +369,55 @@ func buildDeploymentObject(spec *DeploymentSpec) (*reconciler.Object, error) {
...
+		return nil, fmt.Errorf("failed to convert object to k8s object")
+	}
+	deploymentObj, ok := k8sObj.Obj.(*appsv1.Deployment)
+	if !ok {
+		return nil, fmt.Errorf("failed to convert meta object to statefulset object")
+	}
+	if err := addVolumeToPodSpec(&deploymentObj.Spec.Template.Spec, spec.Base.CustomVolumes); err != nil {
+		return nil, err
+	}
+	for ind, _ := range deploymentObj.Spec.Template.Spec.InitContainers {
```
wyzhang: nit: same here and below ind=>idx

api/v1alpha1/cdapmaster_types.go
wyzhang: Nit: give some more context about this custom volume in the comment. A sample snippet of how it might be configured would be even better.

controllers/testdata/logs.json
```diff
@@ -172,16 +177,9 @@
                 "readOnly": true
               },
               {
-                "name": "cdap-cm-vol-my-config-map-1",
```
wyzhang, line 175: nit: is this a cleanup ? Cannot tell why it is related to your change

api/v1alpha1/cdapmaster_types.go
wyzhang: nit: instead of calling these custom volumes, they are just additional volumes that are mounted on all services, if that is the case, just call them additionalVolumes. These volumes can PD, projected, or any types of volumes.  

"custom" makes it sound like a modified type of volume.

controllers/testdata/logs.json
dli357: This is unrelated, but the init containers of the logging service should not have this volume mount. This likely slipped through (and still passes) because we are checking for the expected JSON template to be a superset of the output JSON, when in fact we should be checking the opposite for our integration tests. I've gone ahead and opened [CDAP-18732](https://cdap.atlassian.net/browse/CDAP-18732) to clean this up in the future.

controllers/deployment.go
dli357: Fixed

controllers/deployment.go
dli357: Fixed

controllers/deployment.go
dli357: Fixed

api/v1alpha1/cdapmaster_types.go
dli357: Switched to "AdditionalVolumes" and "AdditionalVolumeMounts" and added an additional comment pointing to the Kubernetes documentation on volumes.

Regarding empty lists, it is fine to omit the list entirely; the JSON or YAML deserialization should properly handle the null list case.

go.mod
```diff
@@ -3,12 +3,10 @@ module cdap.io/cdap-operator
 go 1.12
 
 require (
-	github.com/coreos/etcd v3.3.15+incompatible // indirect
 	github.com/go-logr/logr v0.1.0
 	github.com/nsf/jsondiff v0.0.0-20190712045011-8443391ee9b6
 	github.com/onsi/ginkgo v1.11.0
 	github.com/onsi/gomega v1.8.1
-	github.com/stretchr/testify v1.4.0
```
wyzhang, line 11: nit: are these changes needed?

wyzhang (approved):

go.mod
dli357: These changes were included because I ran `go mod tidy`, so they should no longer be necessary anymore.

PR #77, opened 2022-01-04
[CDAP-18715] Add service running in a sidecar container that polls JMX server on localhost to collect and publish system metrics

ghost:
Changes to enable running a new service that collects system metrics as a sidecar container in stateful sets. PR for CDAP changes that adds the service is under review [here](https://github.com/cdapio/cdap/pull/13766) & [here](https://github.com/cdapio/cdap/pull/13900).


api/v1alpha1/cdapmaster_types.go
```diff
@@ -124,6 +129,9 @@ type CDAPServiceSpec struct {
 	SecretVolumes map[string]string `json:"secretVolumes,omitempty"`
 	// SecurityContext overrides the security context for the service pods.
 	SecurityContext *SecurityContext `json:"securityContext,omitempty"`
+	// DisableSystemMetricsSidecar allows service to disable system metrics sidecar container
+	// even if SystemMetricsExporter spec is present.
+	DisableSystemMetricsSidecar *bool `json:"disableSystemMetricsSidecar,omitempty"`
```
greeshmaswaminathan: Better to make it "EnableMetricCollection" that can default to true.

api/v1alpha1/cdapmaster_types.go
```diff
@@ -210,6 +218,11 @@ type SupportBundleSpec struct {
 	CDAPStatefulServiceSpec `json:",inline"`
 }
 
+//  SystemMetricExporterServiceSpec defines the specification for the SystemMetricsExporter service.
+type SystemMetricExporterServiceSpec struct {
+	CDAPScalableServiceSpec `json:",inline"`
```
greeshmaswaminathan: minor: Do we need the complete service spec here?

config/rbac/role.yaml
```diff
@@ -1,3 +1,4 @@
+
```
greeshmaswaminathan: Remove unchanged file

controllers/cdapmaster/cdapmaster_controller.go
```diff
@@ -69,6 +69,9 @@ type UserInterface struct{}
 // SupportBundle - interface to handle cdapmaster
 type SupportBundle struct{}
 
+// SystemMetricsExporter - interface to handle cdapmaster
+type SystemMetricsExporter struct{}
```
greeshmaswaminathan: Do we need to add new services here? My understanding was that this was intended to support CDF installations with old operator.

controllers/deployment.go
```diff
@@ -40,15 +40,18 @@ type NetworkServiceName = string
 func (d *DeploymentPlan) Init() {
 	d.planMap = make(map[int32]ServiceGroups)
 
-	// Default: each service runs in its own Pod
+	// Default: each service group runs in its own Pod and each service in it's own container.
+	// If there are mltiple services in a pod,
+	// the first one (index 0) will be considered as the
+	// main container and subsequent ones as sidecar containers.
 	d.planMap[0] = ServiceGroups{
 		stateful: map[ServiceGroupName]ServiceGroup{
```
greeshmaswaminathan: May be use a ServiceGroupInfo struct here that has the ServiceGroup and a boolean for collecting metrics.

controllers/deployment.go
```diff
@@ -172,8 +175,17 @@ func buildStatefulSets(master *v1alpha1.CDAPMaster, name string, services Servic
 	spec = spec.withInitContainer(
 		newContainerSpec(master, "StorageInit", dataDir).setArgs(containerStorageMain))
 
+	metricsSidecarEnabled := true
```
greeshmaswaminathan: The ServiceGroupInfo should already have the info for whether metrics collection is enabled, only additionally check if individual service spec has overwritten this value.

greeshmaswaminathan: Also, it may be better to make this check after fetching the CDAPServiceSpec inside the for loop.

controllers/deployment.go
```diff
@@ -185,12 +197,22 @@ func buildStatefulSets(master *v1alpha1.CDAPMaster, name string, services Servic
 		}
 		env := addJavaMaxHeapEnvIfNotPresent(ss.Env, ss.Resources)
 		c := newContainerSpec(master, s, dataDir).setResources(ss.Resources).setEnv(env)
+		// Only main container should run a jmx server if enabed
+		if !isSidecar && metricsSidecarEnabled {
```
greeshmaswaminathan: Add sidecar outside of this loop if metrics collection is enabled. That way the additional check for sidecar can be avoided.

greeshmaswaminathan: Where is the SystemMetricExporterServiceSpec being used ?

config/rbac/role.yaml
ghost: Acknowledged

api/v1alpha1/cdapmaster_types.go
ghost: We need properties for resource requests and limits. Other properties such as ServiceAccount, Env, Configmaps may be required in the future. Changed to inherit from CDAPServiceSpec as replica count wasn't useful here.

controllers/deployment.go
ghost: Acknowledged

controllers/deployment.go
ghost: Added inside loop when service has system metrics enabled. I think since users can control which services to enable metrics for, they can use CR to ensure only one service should have this enabled per pod. Presently we have one service per pod, so such a situation should not arise. For adding container outside the loop, we'll also have code for this in 3 parts:

1. Before loop, initialising boolean to false
2. Inside loop, checking and setting to true for a service if required.
3. After loop, adding the container to sts spec.

So, i think having it in one place, inside the loop, seems better. Have simplified code according to review comments, so it should look cleaner now.

controllers/cdapmaster/cdapmaster_controller.go
ghost: Reverted the commit for this.

api/v1alpha1/cdapmaster_types.go
ghost: I've assumed missing/null as true as a way to have "default true" behaviour.  https://book.kubebuilder.io/reference/markers/crd-validation.html#crd-validation describes an annotation for setting default, but we'll need to update k8s api being used from v1beta1 to v1.
If the SystemMetricsExporter spec isn't present in CR, this flag will not have any effect and no sidecars will be launched. This has the benefit of not starting sidecars in existing instances with old CDAP images and failing.

greeshmaswaminathan: While fetching the instance of CDAPServiceSpec in getCDAPServiceSpec() function, we can set this field to true (if absent) when SystemMetricsExporter is present. That will allow make the check more straight forward while adding side car.

controllers/deployment.go
```diff
@@ -40,7 +40,10 @@ type NetworkServiceName = string
 func (d *DeploymentPlan) Init() {
 	d.planMap = make(map[int32]ServiceGroups)
 
-	// Default: each service runs in its own Pod
+	// Default: each service group runs in its own Pod and each service in it's own container.
```
greeshmaswaminathan: nit: update comment

controllers/testdata/messaging.json
```diff
@@ -319,4 +319,4 @@
     "updateRevision": "cdap-test-messaging-68cf487786",
     "updatedReplicas": 1
   }
-}
\ No newline at end of file
+}
```
greeshmaswaminathan, line 347: nit: revert unchanged file

ghost: There was a missing empty line at the end of the file which my editor added. I think this is required.

controllers/deployment.go
ghost: Removed changes from that line

api/v1alpha1/cdapmaster_types.go
ghost: I've moved the logic to default enableSystemMetrics and check whether sidecar is enabled into the function that was adding the sidecar and simplified the condition here. Renamed to function to `addSystemMetricsServiceIfEnabled`. 
I think it's not expected behaviour that getCDAPServiceSpec will also set a default value. What do you think about the changes?

api/v1alpha1/cdapmaster_types.go
```diff
@@ -91,6 +91,11 @@ type CDAPMasterSpec struct {
 	// To disable this service: either omit or set the field to nil
 	// To enable this service: set it to a pointer to a AuthenticationSpec struct (can be an empty struct)
 	Authentication *AuthenticationSpec `json:"authentication,omitempty"`
+	// SystemMetricsExporter is specification for the CDAP SystemMetricsExporter service that runs as a sidecar container in various pods.
+	// This is an optional service and may not be required for CDAP to be operational.
+	// To disable this service: either omit or set the field to nil
+	// To enable this service: set it to a pointer to a SystemMetricsExporterSpec struct (can be an empty struct)
```
greeshmaswaminathan: nit:  Add a little more info like  - "Adding this will cause metrics collection containers to be created for the services. Use `CDAPServiceSpec.EnableSystemMetrics` field to disable this container creation for specific services.

api/v1alpha1/cdapmaster_types.go
```diff
@@ -136,6 +141,10 @@ type CDAPServiceSpec struct {
 	AdditionalVolumeMounts []corev1.VolumeMount `json:"additionalVolumeMounts,omitempty"`
 	// SecurityContext overrides the security context for the service pods.
 	SecurityContext *SecurityContext `json:"securityContext,omitempty"`
+	// EnableSystemMetrics allows service to disable system metrics service which runs in a sidecar container
+	// by setting it to false. When property is not specified, taken as true
+	// SystemMetricsExporter spec should also be present in CR.
+	EnableSystemMetrics *bool `json:"enableSystemMetrics,omitempty"`
```
greeshmaswaminathan, line 152: nit: Add a little more info like - "Optional field that is considered along with CDAPMasterSpec.SystemMetricsExporter to start a metric collection container in services. SystemMetricsExporter is a global setting in the CDAPMasterSpec when present enables metric collection for all services. For overriding this value (prevent creating metric collection container) for individual services, this field can be set as false. No effect when SystemMetricsExporter is not present. "

greeshmaswaminathan (approved):

api/v1alpha1/cdapmaster_types.go
ghost: Added more info

api/v1alpha1/cdapmaster_types.go
ghost: Added

controllers/deployment.go
```diff
@@ -226,6 +230,29 @@ func buildStatefulSets(master *v1alpha1.CDAPMaster, name string, services Servic
...
 
+// addSystemMetricsServiceIfEnabled adds a sidecar container for
+// SystemMetricsExporterService if enabled in service spec.
+// Returns an error and a boolean representing whether the sidecar was added
+func addSystemMetricsServiceIfEnabled(stsSpec *StatefulSpec, master *v1alpha1.CDAPMaster,
+	service *v1alpha1.CDAPServiceSpec, dataDir string) (err error, added bool) {
+	if master.Spec.SystemMetricsExporter == nil || service == nil {
+		return nil, false
+	}
+	if service.EnableSystemMetrics == nil {
```
dli357, line 237: Should we be enabling system metrics by default? It seems like a cleaner way to implement this would be to disable it by default and only apply it to services that have it explicitly enabled.

api/v1alpha1/cdapmaster_types.go
```diff
@@ -225,6 +238,11 @@ type SupportBundleSpec struct {
 	CDAPStatefulServiceSpec `json:",inline"`
 }
 
+//  SystemMetricExporterServiceSpec defines the specification for the SystemMetricsExporter service.
+type SystemMetricExporterServiceSpec struct {
```
dli357: nit: Rename to just `SystemMetricExporterSpec`, thanks!

controllers/constants.go
```diff
@@ -106,6 +110,11 @@ const (
 	javaReservedNonHeap       = int64(768 * 1024 * 1024)
 	javaMaxHeapSizeEnvVarName = "JAVA_HEAPMAX"
 
+	// System Metrics sidecar related
+	defaultJMXport            = 11022
+	javaOptsEnvVarName        = "OPTS"
+	runJMXServerJavaOptFormat = "-Dcom.sun.management.jmxremote=true  -Dcom.sun.management.jmxremote.port=%s  -Dcom.sun.management.jmxremote.ssl=false  -Dcom.sun.management.jmxremote.authenticate=false"
```
dli357: This represents a potential security vulnerability as anybody who can connect to the JMX server now has remote control (and likely remote code execution) to the Java process. If we are to let this PR through, can we (at minimum) open a tracking Jira to enable SSL client verification for JMX?

controllers/spec.go
```diff
@@ -84,6 +84,14 @@ func (s *ContainerSpec) setArgs(arg ...string) *ContainerSpec {
 }
 
 func (s *ContainerSpec) addEnv(name, value string) *ContainerSpec {
+	// find existing EnvVar with same name
```
dli357: This seems counterintuitive to general logic of adding an environment variable with an existing key. Since only the Java OPTS environment variable requires appending to an existing variable, can we make that logic specific to the system metrics service?

api/v1alpha1/cdapmaster_types.go
ghost: Done

controllers/spec.go
ghost: Yes, removed changes from here and added to function responsible for adding system metrics sidecar.

controllers/constants.go
ghost: Yes you're right! I tried connecting to jmx server port of messaging pod (or any other pod in the same cluster) from the logs pod using telnet and was able to connect. The output of running netstat shows that jmx port (11022) is bound to 0.0.0.0 and can therefore connect to requests from external IPs.
Screenshot 2022-01-27 at 2 48 55 PM
I found [this](https://stackoverflow.com/questions/35367232/how-to-configure-jmx-to-bind-to-localhost-only) question on stackoverflow about binding jmx server to localhost. Following the comment on the answer, I added an extra flag `-Dcom.sun.management.jmxremote.host=localhost`. After this I was no longer able to connect to jmx servers running on other pods. Netstat shows that 11022 is bound to 127.0.0.1 as required.
Screenshot 2022-01-27 at 2 39 22 PM
With this change, only requests from localhost will be accepted by the JMX server. Since the metrics collection service runs as a sidecar container, it is able to communicate with the jmx server over localhost. The service which polls the JMX server doesn't handle any user requests (no external API exposed). I think the only way to exploit the unauthenticated JMX server now would be to have access to a terminal on the pod. Is RCE still a major concern? Can you give an estimate for the amount work required to add ssl authentication?

dli357: I think it would be acceptable for us to limit it to localhost for now. Does the JMX server run on any pods which run user code? There's a minor concern in the future of the JMX server allowing for local sandbox escapes (for example, if we implement a Java sandbox around user code within the primary user code container, then malicious code could leverage this to escape from that sandbox).

Regarding the SSL authentication question, we already have some tentative plans for enabling client and server-side validation of SSL certificates; In that situation, we would simply just need to pass the trusted public key certificate into the secret volume and use that to authenticate the caller (sidecar container) which should have access to the private key. However, that hasn't been setup yet, and currently the public and private keys are part of the same SSL certificate, so it's a nontrivial amount of work to complete immediately.

controllers/deployment.go
ghost: Acknowledged. Changed so that sidecar is enabled only when SystemMetricsExporter is present in Master spec and this flag is set to true.

controllers/constants.go
ghost: We plan to run the server on all Statefulsets, and from conversations with the team, I think Appfabric runs user code when RBAC isn't enabled. I don't understand how user code is sandboxed well enough to comment on the possibility of JMX server being exploited. I've opened a feature improvement on [Jira](https://cdap.atlassian.net/browse/CDAP-18783) for adding SSL authentication.

dli357: Sounds good, thanks!

dli357 (approved):
LGTM, thanks!

ghost:
Thanks for the reviews!

PR #79, opened 2022-01-12
Add container lifecycle hooks in CDAPServiceSpec

rahul1995:


ghost (changes requested):
Changes look good, we should have unit tests that cover these changes too. I think we'll need to modify /testdata for that.

rahul1995:
Thanks for the review @arjansbal . Updated the test-case.

ghost (approved):
LGTM

api/v1alpha1/cdapmaster_types.go
```diff
@@ -124,6 +124,8 @@ type CDAPServiceSpec struct {
 	SecretVolumes map[string]string `json:"secretVolumes,omitempty"`
 	// SecurityContext overrides the security context for the service pods.
 	SecurityContext *SecurityContext `json:"securityContext,omitempty"`
+	// Lifecycle is to specify Container Lifecycle hooks provided by Kubernetes for containers
```
dli357: Please add a comment specifying how these lifecycle hooks are applied to containers, e.g. are they applied to all containers (including init containers)? Or only specific containers?

templates/cdap-deployment.yaml
```diff
@@ -88,6 +88,33 @@ spec:
           {{if $c.ImagePullPolicy}}
           imagePullPolicy: {{$c.ImagePullPolicy}}
           {{end}}
+          {{if $c.Lifecycle}}
```
dli357: Can we directly serialize the `Lifecycle` object instead of manually copying individual components? Otherwise you end up with a lot of nested if statements. For an example, please see https://github.com/cdapio/cdap-operator/pull/73, specifically [deployment.go#341](https://github.com/cdapio/cdap-operator/blob/fdf0de41674e41b537ffa186b006d2407a61a31a/controllers/deployment.go#L341) and [deployment.go#372](https://github.com/cdapio/cdap-operator/blob/fdf0de41674e41b537ffa186b006d2407a61a31a/controllers/deployment.go#L372).

templates/cdap-sts.yaml
```diff
@@ -126,6 +126,33 @@ spec:
           {{if $c.ImagePullPolicy}}
           imagePullPolicy: {{$c.ImagePullPolicy}}
           {{end}}
+          {{if $c.Lifecycle}}
```
dli357: Same as above

dli357 (changes requested):

templates/cdap-deployment.yaml
rahul1995: Thanks for the suggestion. Updated.

api/v1alpha1/cdapmaster_types.go
rahul1995: Updated.

templates/cdap-sts.yaml
rahul1995: Updated.

templates/cdap-deployment.yaml
```diff
@@ -11,6 +11,12 @@
 # WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
 # License for the specific language governing permissions and limitations under
 # the License.
+#
+#
+# NOTE: Some properties are highly nested and it is not feasible to pick them from the template file
+# hence those are added directly in the code.
+# For example, please check the method "../controllers/deployment.go#buildDeploymentObject" and see
+# how container's lifecycle property is added to the k8s object
```
dli357: nit: period at end.

templates/cdap-sts.yaml
```diff
@@ -11,6 +11,12 @@
 # WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
 # License for the specific language governing permissions and limitations under
 # the License.
+#
+#
+# NOTE: Some properties are highly nested and it is not feasible to pick them from the template file
+# hence those are added directly in the code.
+# For example, please check the method "../controllers/deployment.go#buildStatefulSetsObject" and see
+# how container's lifecycle property is added to the k8s object
```
dli357: nit: period at end.

dli357 (approved):
LGTM, please don't forget to rebase. Thanks!

templates/cdap-deployment.yaml
rahul1995: Done.

templates/cdap-sts.yaml
rahul1995: Fixed.

PR #80, opened 2022-01-14
add new podGroup struct that supports sidecars

ghost:


ghost:
No longer required in new design for metrics sidecar.

PR #81, opened 2022-01-20
[CDAP-18752] Dependent Objects not deleted on CR deletion for GKE version 1.20+

ghost:
[CDAP 18752](https://cdap.atlassian.net/browse/CDAP-18752)
Since K8s version 1.20, the parent references aren't being used to enforce cascade deletion of objects(dependents like STS, deployments and Configmap) on CR deletion. K8s isn't picking up the owner references set by operator library we use.  The controller-reconciler library sets the owner refs for us in the format:
```
ownerReferences:
  - apiVersion: cdap.cdap.io/v1alpha1
    blockOwnerDeletion: true
    controller: true
    kind: '*v1alpha1.CDAPMaster'
    name: cdap
    uid: 5ecd0cab-2b6a-4898-b03b-ac8cb6784762
  ```
  Changing the `kind` field to `CDAPMaster` fixes the issue in version 1.20+.  
  Changes in this PR:

1. Change the format set by genericreconciler
2. Make sure we replace existing owner refs (using old format) with the correct format. The current behaviour of the library is to [append new owner refs](https://github.com/cdapio/cdap-operator/blob/develop/vendor/sigs.k8s.io/controller-reconciler/pkg/reconciler/manager/k8s/manager.go#L125) instead of replacing them. This causes reconciling to get stuck as k8s doesn't allow more than 1 owner refs with field `controller: true`. This change detects when the new owner ref is the updated version of an existing owner ref and replaces old ref instead of appending.

Manually tested changes on GKE cluster with version 1.21 and 1.19. Replaced existing operator with changes from this PR. Was able to delete CR without dependent Objects being left behind.

### DIFF
--- a/vendor/sigs.k8s.io/controller-reconciler/pkg/reconciler/manager/k8s/manager.go
+++ b/vendor/sigs.k8s.io/controller-reconciler/pkg/reconciler/manager/k8s/manager.go
@@ -19,6 +19,12 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"log"
+	"os"
+	"reflect"
+	"strings"
+	"text/template"
+
 	corev1 "k8s.io/api/core/v1"
 	policyv1 "k8s.io/api/policy/v1beta1"
 	"k8s.io/apimachinery/pkg/api/meta"
@@ -29,14 +35,9 @@ import (
 	"k8s.io/apimachinery/pkg/util/validation/field"
 	"k8s.io/apimachinery/pkg/util/yaml"
 	"k8s.io/client-go/kubernetes/scheme"
-	"log"
-	"os"
-	"reflect"
 	"sigs.k8s.io/controller-reconciler/pkg/reconciler"
 	"sigs.k8s.io/controller-reconciler/pkg/reconciler/manager"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	"strings"
-	"text/template"
 )
 
 // constants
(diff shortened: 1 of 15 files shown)

### CODE REVIEW

**Review by chtyim — APPROVED** [review]

File: vendor/sigs.k8s.io/controller-reconciler/pkg/reconciler/manager/k8s/manager.go [hunk 1]
```diff
@@ -116,6 +117,23 @@ func isReferringSameObject(a, b metav1.OwnerReference) bool {
 	return aGV == bGV && a.Kind == b.Kind && a.Name == b.Name
 }
 
+func isKindFormatUpdate(a, b metav1.OwnerReference) bool {
```
wyzhang: Not sure if I understand what this function is trying to do. How do we know b is using old format for the kind?

wyzhang: Rename this to needUpdateOwnerRefKindFormat(). 
Add comments to give context on old vs new format for GKE 1.19 vs 1.20+. Explain what returning true mean and what the call should do when get back a true.

vendor/sigs.k8s.io/controller-reconciler/pkg/reconciler/manager/k8s/manager.go
```diff
@@ -116,6 +117,23 @@ func isReferringSameObject(a, b metav1.OwnerReference) bool {
...
+	if a.APIVersion != b.APIVersion ||
+		a.Name != b.Name ||
+		a.UID != b.UID ||
+		getKindFromOldFormat(b.Kind) != a.Kind {
```
wyzhang: // No need to update if these are two different kind 
if  ... !isKindEquivalent(a.kind, b.kind) {
 return false
}

if (a.kind != b.kind) {
return true;
}

return false

wyzhang (approved):

vendor/sigs.k8s.io/controller-reconciler/pkg/reconciler/manager/k8s/manager.go
ghost: Acknowledged

vendor/sigs.k8s.io/controller-reconciler/pkg/reconciler/manager/k8s/manager.go
ghost: Acknowledged

PR #82, opened 2022-01-20
[CDAP-18752] Dependent Objects not deleted on CR deletion for GKE version 1.20+ without changing vendor code

ghost:
[CDAP 18752](https://cdap.atlassian.net/browse/CDAP-18752)
Since K8s version 1.20, the parent references aren't being used to enforce cascade deletion of objects(dependents like STS, deployments and Configmap) on CR deletion. K8s isn't picking up the owner references set by operator library we use.  The controller-reconciler library sets the owner refs for us in the format:
```
ownerReferences:
  - apiVersion: cdap.cdap.io/v1alpha1
    blockOwnerDeletion: true
    controller: true
    kind: '*v1alpha1.CDAPMaster'
    name: cdap
    uid: 5ecd0cab-2b6a-4898-b03b-ac8cb6784762
  ```
Changing the `kind` field to `CDAPMaster` fixes the issue in version 1.20+, however that requires changes in vendor directory and may cause issues while upgrading dependency versions.
Changes in this PR:

1. Use a field provided in reconciler.Object to mark dependent objects for deletion. This doesn't fix the main issue but ensures resources are deleted by generic reconciler [here](https://github.com/cdapio/cdap-operator/blob/develop/vendor/sigs.k8s.io/controller-reconciler/pkg/genericreconciler/genericreconciler.go#L195).

Manually tested changes on GKE cluster with version 1.21 and 1.19. Replaced existing operator with changes from this PR. Was able to delete CR without dependent Objects being left behind.

### DIFF
--- a/controllers/cdapmaster_controller.go
+++ b/controllers/cdapmaster_controller.go
@@ -173,6 +173,10 @@ func (h *ConfigMapHandler) Observables(rsrc interface{}, labels map[string]strin
 func (h *ConfigMapHandler) Objects(rsrc interface{}, rsrclabels map[string]string, observed, dependent, aggregated []reconciler.Object) ([]reconciler.Object, error) {
 	var expected []reconciler.Object
 	m := rsrc.(*v1alpha1.CDAPMaster)
+	if m.DeletionTimestamp != nil {
+		markObjectsForDeletion(observed)
+		return expected, nil
+	}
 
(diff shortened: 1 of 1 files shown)


ghost:
Closing PR as fix was made in vendor code in [PR](https://github.com/cdapio/cdap-operator/pull/81).

PR #85, opened 2022-03-08
Add support for tethering services

rmstar:
Add tetheringAgent and artifactCache services.

Tethering Agent Service runs on CDAP instance that is initiating tethering with a peer. It's responsible for establishing a tethering connection and polls the peer for commands to run (start pipeline, stop pipeline). It needs to be stateful because it writes to TMS.

Artifact Cache Service is responsible for caching artifacts belonging to tethered CDAP instances when the local CDAP is running programs on behalf of a tethered peer. It needs to be stateful because it caches artifacts locally.


api/v1alpha1/cdapmaster_types.go
```diff
@@ -81,6 +81,16 @@ type CDAPMasterSpec struct {
 	// To disable this service: either omit or set the field to nil
 	// To enable this service: set it to a pointer to a SupportBundleSpec struct (can be an empty struct)
 	SupportBundle *SupportBundleSpec `json:"supportBundle,omitempty"`
+	// TetheringAgent is specification for the CDAP Tethering Agent service.
+	// This is an optional service and may not be required for CDAP to be operational.
+	// To disable this service: either omit or set the field to nil
+	// To enable this service: set it to a pointer to a TetheringAgentSpec struct (can be an empty struct)
+	TetheringAgent *TetheringAgentSpec `json:"tetheringAgent,omitempty"`
```
wyzhang, line 88: nit: the comment doesn't provide more details than the code. 

Can you clarify on 
why need tethering agent?
at the high level what it does?
why it needs to be stateful?

dli357: +1 on why it needs to be stateful.

rmstar: I have updated the PR description with these details. 
Regarding why it needs to be stateful:
Artifact Cache: storage needed to cache tethered peers' artifacts
Tethering Agent: reads/writes to TMS.

dli357 (approved):

PR #86, opened 2022-03-31
[CDAP-18995] Add env var support across all CDAP services

dli357:
This PR adds support for an `env` field in the CDAPMasterSpec which applies a set of environment variables across all CDAP master services. This helps reduce configuration duplication.


controllers/deployment.go
```diff
@@ -343,9 +356,38 @@ func buildDeployment(master *v1alpha1.CDAPMaster, name string, services ServiceG
...
+				}
+				// Overwrite any conflicting environment variables with the service-specific values
+				env[ind] = envVar
+				overwritten = true
+			}
+		}
+		if !overwritten {
+			env = append(env, envVar)
+		}
+	}
```
ghost: This loops seem to be doing a couple of things which we should separate out to make it easier to understand:
1) Detecting duplicate env vars in MasterSpec and Service Spec. This logic can be put in a separate function which can be called for both specs.
2) Merging two lists while giving priority to service spec when the same env var is present in both specs. 

For 2 we can create an empty map[string]string or map[string]corev1.envvar and do the following:
1) Insert all env vars from master spec as key value pairs with var.Name as the key.
2) Repeat step 1 for service spec, this should overwrite same keys inserted in step 1.
3) Convert this map into a list of env vars.

This should make the code easier to follow.

controllers/deployment.go
```diff
@@ -186,9 +186,15 @@ func buildStatefulSets(master *v1alpha1.CDAPMaster, name string, services Servic
 			continue
 		}
 
-		c := serviceContainerSpec(ss, master, dataDir, s)
+		// Merge environment variables between service spec and master spec
+		env, err := mergeEnvVars(&master.Spec, ss)
```
ghost: We pass master and service spec (ss) to the serviceContainerSpec function, we could move this merging step into serviceContainerSpec() and reduce duplication while creating STS spec and Deployment Spec.

controllers/deployment.go
ghost: Another way which generalises both steps from previous comment into 1 is to create a helper function:
```
func insertIntoMap(m map[string]string, vals []corev1.envVar, allowOverwrite) err {
  // insert vals into m
  // return error if allowOverwrite is false and key already exists in m
}
```
With this, we can simply this loop into the following calls:
1) check insertIntoMap(make(map[string]string, masterSpec.env, false) == nil // no duplicate keys
2) check insertIntoMap(make(map[string]string, serviceSpec.env, false) == nil // no duplicate keys
3) 
```
m = make(map[string]string
insertIntoMap(m, masterSpec.env, true)
insertIntoMap(m, serviceSpec.env, true)
```
4) Convert m to an array and return.

controllers/deployment.go
```diff
@@ -186,9 +186,15 @@ func buildStatefulSets(master *v1alpha1.CDAPMaster, name string, services Servic
...
+		if err != nil {
+			return nil, err
+		}
+
+		c := serviceContainerSpec(ss, master, dataDir, s, env)
 		spec = spec.withContainer(c)
-		if err := addSystemMetricsServiceIfEnabled(spec, master, ss, dataDir, c); err != nil {
+		if err := addSystemMetricsServiceIfEnabled(spec, master, ss, dataDir, c, env); err != nil {
```
ghost: I think this will cause the env vars for the main service to be used for the sidecar service also while ignoring the env vars of the sidecar service. If this is the case, it's undesirable as we will not be able to reduce the java heap size for the sidecar independently of the main service. One fix would be to re generate env while using systemMetricsExporter spec (see line #245). 
A cleaner solution IMO would be to add the mergeEnvVars call to serviceContainerSpec() as mentioned in a comment above.

controllers/deployment.go
dli357: Fixed, I used a double map to do this, should be more clear now. Thanks!

controllers/deployment.go
dli357: Fixed, thanks!

controllers/deployment.go
dli357: Fixed, seems like the original pattern of passing an env var into the serviceContainerSpec function came from an earlier prototype, and I didn't simplify. Thanks!

controllers/deployment.go
```diff
@@ -343,14 +354,56 @@ func buildDeployment(master *v1alpha1.CDAPMaster, name string, services ServiceG
...
+		seenVars[envVar.Name] = true
+		envMap[envVar.Name] = envVar
+	}
+
+	// Convert the map to a sorted slice.
+	env := []corev1.EnvVar{}
+	for _, envVar := range envMap {
+		env = append(env, envVar)
+	}
+	sort.Sort(ByEnvKey(env))
```
ghost, line 390: nit: Why do we need to sort?

dli357: Consistent ordering seems like it'll be nicer to have (to spot errors, etc), and we could end up with flaky tests if we don't sort as well. Performance loss should be minimal since we're unlikely to be adding a huge amount of environment variables, but if it does become a problem, we can also just move the sort function to the test itself.

controllers/deployment.go
```diff
@@ -343,14 +354,56 @@ func buildDeployment(master *v1alpha1.CDAPMaster, name string, services ServiceG
...
 }
 
+type ByEnvKey []corev1.EnvVar
+
+func (k ByEnvKey) Len() int           { return len(k) }
+func (k ByEnvKey) Swap(i, j int)      { k[i], k[j] = k[j], k[i] }
+func (k ByEnvKey) Less(i, j int) bool { return k[i].Name < k[j].Name }
+
+func mergeEnvVars(baseEnvVars []corev1.EnvVar, overwriteEnvVars []corev1.EnvVar) ([]corev1.EnvVar, error) {
+	// Merge master and service-specific environment variables
```
ghost: Nit: function parameters are named base & overwrite, so maybe we should follow the same naming in comments also?

ghost (approved):
LGTM. Remember to squash commits before merging.

controllers/deployment.go
dli357: Done

PR #91, opened 2022-09-22
Adding replicas support to the Runtime service

chtyim:
- Added a new CDAPScalableStatefulSpec type to have the replicas field
- Refactoring of the utils.go to reduce code duplication


controllers/deployment.go
```diff
@@ -691,6 +691,24 @@ func getReplicas(master *v1alpha1.CDAPMaster, services ServiceGroup) (int32, err
 	return replicas, nil
 }
 
+// getServiceReplicas returns the Replicas of a scalable service. If the service is not scalable, return 1 as the replicas.
```
wyzhang: nit: "return 1 as the replica count"

controllers/utils.go
```diff
@@ -56,100 +52,128 @@ func getCDAPServiceSpec(master *v1alpha1.CDAPMaster, service ServiceName) (*v1al
...
-	if !ok {
-		return nil, fmt.Errorf("failed to cast to poiter to %v for %v", fieldNameCDAPExternalServiceSpec, service)
-	}
-	return ret, nil
 }
 
-func min(x, y int64) int64 {
-	if x < y {
-		return x
+// getFieldValue finds a field value in the given struct object that passes the predicate and return it as a *reflect.Value
```
wyzhang: nit: the comment should call out the field to be found can be in nested structure and only the value of the first matching field name will be returned.

wyzhang (approved):

controllers/deployment.go
```diff
@@ -671,17 +676,12 @@ func getSecurityContext(master *v1alpha1.CDAPMaster, services ServiceGroup) (*v1
...
-			if replicas == 0 {
-				replicas = *spec.Replicas
-			} else if replicas != *spec.Replicas {
-				return 0, fmt.Errorf("value of field Replicas not the same across (%s)", strings.Join(services, ","))
-			}
+		if serviceReplicas, err := getServiceReplicas(master, service); err != nil {
+			return 0, err
+		} else if replicas == 0 {
+			replicas = serviceReplicas
+		} else if replicas != serviceReplicas {
```
kapadiap, line 683: just curious: why do we require all services have same number of replicas?

kapadiap (approved):

controllers/deployment.go
chtyim: I think this was a side effect of allowing running multiple CDAP services in the same pod. The Replicas setting in the CDAP CR is per service, hence if multiple services are running in the same pod, they need to have the same replicas setting in the CR.

controllers/utils.go
chtyim: Updated

controllers/deployment.go
chtyim: Updated. Also added the type check in the predicate to make sure the type is `*int32`

masoud-io (approved):

PR #93, opened 2022-10-10
[CDAP-19678] Add support for affinity in CDAP service spec

arjan-bal:
[CDAP-19678](https://cdap.atlassian.net/browse/CDAP-19678)
[Pod Affinity/Anti Affinity](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#affinity-and-anti-affinityhttps://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#affinity-and-anti-affinity) allow us to tell the k8s scheduler how to to schedule pods on different nodes.
For CDAP, pod anti affinity can be used to ensure services that have high resource usage are scheduled on different nodes so that they can scale more freely.


rahul1995 (approved):
LGTM

(Not scoped for this PR) I feel we should embed `BaseSpec` type into `DeploymentSpec` and `StatefulSpec`.
Currently every method (for e.g. `SetAffinity`) in BaseSpec needs to be replicated in these two child specs which can be avoided if the Base was embedded in them.

arjan-bal:
> LGTM
> 
> (Not scoped for this PR) I feel we should embed `BaseSpec` type into `DeploymentSpec` and `StatefulSpec`. Currently every method (for e.g. `SetAffinity`) in BaseSpec needs to be replicated in these two child specs which can be avoided if the Base was embedded in them.

I tried to embed BaseSpec inside StsSpec and DeploymentSpec, but ran into the following issue:
```
spec := newStatefulSpec(master, objName, labels, cconf, hconf, sysappconf).
        setServiceAccountName(serviceAccount).
        setNodeSelector(nodeSelector).
        setRuntimeClassName(runtimeClass).
        setPriorityClassName(priorityClass).
        setSecurityContext(securityContext).
        setReplicas(replicas)

    spec = spec.(*StatefulSpec) // invalid operation: spec (variable of type *BaseSpec) is not an interface

```
We have such chaining of functions in deployment.go. If the shared methods are defined on BaseSpec, the return type will be BaseSpec.  We can't caste BaseSpec back to StatefulSetSpec because BaseSpec isn't an interface. This will require some more thought. We may need to introduce an interface.

masoud-io (approved):

PR #94, opened 2022-10-18
[CDAP-19803] Migrate CDAP operator to Kubebuilder v3

arjan-bal:
[CDAP-19803](https://cdap.atlassian.net/browse/CDAP-19803) 
Migration guide: [Kubebuilder docs](https://cdap.atlassian.net/browse/CDAP-19803)

Subtasks:
1. Upgrade Go version to 1.19 which is required by dependencies.
2. Upgrade Github workflow to use Go v1.19 and pull newer version of Kubebuilder.
3. Tested creation, deletion, updating of CR in a GKE cluster.
4. Updated Go version in Dockerfile.
5. Rename vendor folder to vendor_old as `vendor` directory is used by default for vendoring by Go modules.  Partial vendoring of dependencies isn't supported. See [this discussion](https://groups.google.com/g/golang-nuts/c/P_Whhd2qFhc).
6. Replace `runtime.Object` variables type to `client.Object` in controller-reconciler. This is required due to a breaking change introduced in controller-runtime 0.7.0 ([Change log](https://github.com/kubernetes-sigs/controller-runtime/releases/tag/v0.7.0)).


rahul1995 (approved):

Makefile
```diff
@@ -11,70 +11,123 @@ else
...
+	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
+	$(KUSTOMIZE) build config/default | kubectl apply -f -
+
+.PHONY: undeploy
+undeploy: ## Undeploy controller from the K8s cluster specified in ~/.kube/config. Call with ignore-not-found=true to ignore resource not found errors during deletion.
+	$(KUSTOMIZE) build config/default | kubectl delete --ignore-not-found=$(ignore-not-found) -f -
+
+##@ Build Dependencies
+## Location to install dependencies to
+LOCALBIN ?= $(shell pwd)/bin
```
dli357, line 105: Did this use a hardcoded directory previously?

vendor_old/sigs.k8s.io/controller-reconciler/pkg/genericreconciler/genericreconciler.go
```diff
@@ -68,7 +70,7 @@ func (gr *Reconciler) ReconcileResource(namespacedname types.NamespacedName) (re
 	period := DefaultReconcilePeriod
 	expected := []reconciler.Object{}
 	resource := gr.resource.(runtime.Object).DeepCopyObject()
-	nname := reflect.TypeOf(resource).String() + "/" + namespacedname.String()
+	nname := strings.Trim(reflect.TypeOf(resource).String(), "*") + "/" + namespacedname.String()
```
dli357, line 73: What changed to require this string trimming?

dli357 (commented):
Did we manually patch the genericreconciler library? Is there any updated one that works out-of-the-box for kubebuilder v3?

vendor_old/sigs.k8s.io/controller-reconciler/pkg/genericreconciler/genericreconciler.go
arjan-bal: The line previously resulted in a name like `*v1alpha1.CDAPMaster`. The leading `*`  is present as TypeOf(resource) is a pointer. The `*` doesn't cause any unexpected behaviour presently as it's only used in log statements below. 
New k8s versions may not understand resource names with a leading `*`. Similar issue was fixed in https://github.com/cdapio/cdap-operator/pull/81

Makefile
arjan-bal: > Did we manually patch the genericreconciler library? Is there any updated one that works out-of-the-box for kubebuilder v3?

genericreconciler is owned by Google and seems to be no longer maintained. The last version was released in Nov 2019. The [Reconcile method](https://pkg.go.dev/github.com/kubernetes-sigs/application/pkg/genericreconciler#Reconciler.Reconcile) in the latest version doesn't have the extra `context.Context` parameter which is the change that I added.

arjan-bal:
> Did we manually patch the genericreconciler library? Is there any updated one that works out-of-the-box for kubebuilder v3?

genericreconciler is owned by Google and seems to be no longer maintained. The last version was released in Nov 2019. The [Reconcile method](https://pkg.go.dev/github.com/kubernetes-sigs/application/pkg/genericreconciler#Reconciler.Reconcile) in the latest version doesn't have the extra `context.Context` parameter which is the change that I added.

Makefile
arjan-bal: No it didn't. This is code generated by kubebuilder v3. 
Source template: [Kubebuilder repo](https://github.com/kubernetes-sigs/kubebuilder/blob/c73d4d5faa30f5cf3c510a688cb0db17f5609649/pkg/plugins/golang/v3/scaffolds/internal/templates/makefile.go#L185-L187)
Kubebuilder v3 installs the required versions of `controller-gen`, `kustomize` and `envtest` in a directory inside the cloned repo. I think is done to prevent collisions with different versions of the same dependency present in `GOBIN` or `GOPATH/bin`

dli357 (approved):

PR #101, opened 2023-06-28
[CDAP-20714] Support creation of Router as a LoadBalancer service

arjan-bal:
## [CDAP-20714](https://cdap.atlassian.net/browse/CDAP-20714)

List of fields added:
* LoadBalancerIP: Optional static IP to be used for load balancer services.
* ServiceAnnotations: Used to configure Cloud Provider specific properties for load balancers such as type of load balancer in GCP. The field name `Annotations` is already present in the ObjectMeta struct and is used to set the metadata for the k8s deployment, hence the name `ServiceAnnotations` is used.

Manually tested the operator by making `Router` a load balancer service in a GKE cluster with a reserved static IP.

[CDAP-20714]: https://cdap.atlassian.net/browse/CDAP-20714?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/v1alpha1/cdapmaster_types.go
+++ b/api/v1alpha1/cdapmaster_types.go
@@ -184,8 +184,12 @@ type CDAPExternalServiceSpec struct {
 	CDAPScalableServiceSpec `json:",inline"`
 	// ServiceType is the service type in kubernetes, default is NodePort.
 	ServiceType *string `json:"serviceType,omitempty"`
+	// LoadBalancerIP is the static IP address reserved for the service's load balancer.
+	LoadBalancerIP *string `json:"loadBalancerIP,omitempty"`
 	// ServicePort is the port number for the service.
 	ServicePort *int32 `json:"servicePort,omitempty"`
+	// Annotations are the metadata annotations for the kubernetes service. They can be used to configure cloud provider specific behavior for LoadBalancer services.
+	Annotations map[string]string `json:"Annotations,omitempty"`
 }
 
 // CDAPStatefulServiceSpec defines the base specification for stateful master services.
--- a/controllers/testdata/cdap_master_cr.json
+++ b/controllers/testdata/cdap_master_cr.json
@@ -251,6 +251,11 @@
       "replicas": 2
     },
     "router": {
+      "Annotations": {
+        "networking.gke.io/load-balancer-type":"Internal"
+      },
+      "loadBalancerIP": "172.25.80.9",
+      "serviceType": "LoadBalancer",
       "affinity": {
         "podAntiAffinity": {
           "preferredDuringSchedulingIgnoredDuringExecution": [
--- a/api/v1alpha1/zz_generated.deepcopy.go
+++ b/api/v1alpha1/zz_generated.deepcopy.go
@@ -83,10 +83,22 @@ func (in *CDAPExternalServiceSpec) DeepCopyInto(out *CDAPExternalServiceSpec) {
 		*out = new(string)
 		**out = **in
 	}
+	if in.LoadBalancerIP != nil {
+		in, out := &in.LoadBalancerIP, &out.LoadBalancerIP
+		*out = new(string)
+		**out = **in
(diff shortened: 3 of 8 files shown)


api/v1alpha1/cdapmaster_types.go
```diff
@@ -184,8 +184,12 @@ type CDAPExternalServiceSpec struct {
 	CDAPScalableServiceSpec `json:",inline"`
 	// ServiceType is the service type in kubernetes, default is NodePort.
 	ServiceType *string `json:"serviceType,omitempty"`
+	// LoadBalancerIP is the static IP address reserved for the service's load balancer.
```
chtyim, line 187: Do we need to use preallocated IP? Or can it be auto allocated when the underlying ILB is created?

arjan-bal: This is an optional field, so it can be skipped. If cloud provider supports, we can provide a static IP. This is useful if we want to create VPC routes before the LB is created or we want to ensure the routes remain correct even if the load balancer is re-created.
* [k8s docs](https://kubernetes.io/docs/concepts/services-networking/service/#loadbalancer:~:text=Some%20cloud%20providers%20allow%20you%20to%20specify%20the%20loadBalancerIP)
* [GKE docs](https://cloud.google.com/kubernetes-engine/docs/concepts/service-load-balancer-parameters#spd-static-ip)

arjan-bal: On thinking a bit, this field isn't required from CDF perspective. Should I remove it?

arjan-bal: Discussed offline, it is fine to keep the field even though it is not a requirement for CDF usage.

api/v1alpha1/cdapmaster_types.go
```diff
@@ -184,8 +184,12 @@ type CDAPExternalServiceSpec struct {
...
+	LoadBalancerIP *string `json:"loadBalancerIP,omitempty"`
 	// ServicePort is the port number for the service.
 	ServicePort *int32 `json:"servicePort,omitempty"`
+	// ServiceAnnotations are the metadata annotations for the kubernetes service. They can be used to configure cloud provider specific behavior for LoadBalancer services.
+	ServiceAnnotations map[string]string `json:"serviceAnnotations,omitempty"`
```
chtyim: Just call it annotations, as it is already inside the service object.

controllers/testdata/cdap_master_cr.json
```diff
@@ -251,6 +251,11 @@
       "replicas": 2
     },
     "router": {
+      "serviceAnnotations": {
+        "networking.gke.io/load-balancer-type":"Internal"
+      },
+      "loadBalancerIP": "172.25.80.9",
```
chtyim, line 257: Since IP is option, add a test that doesn't have IP too.

chtyim (approved):
LGTM. Just a couple minor comments.

api/v1alpha1/cdapmaster_types.go
arjan-bal: Changed. It was initially named ServiceAnnotations because a `metav1.ObjectMeta` struct is embedded inside `CDAPServiceSpec` and  calling CDAPServiceSpec.Annotations would give the annotations of the embedded object. After introducing a field named `Annotations` in this struct the annotations in the embedded struct can be accessed using `CDAPServiceSpec.ObjectMeta.Annotations`.

controllers/testdata/cdap_master_cr.json
arjan-bal: There is another service named `userInterface` that doesn't sepecify the `loadBalancerIP` field and the output service spec also doesn't contain the `loadBalancerIP` field.

PR #106, opened 2023-09-27
[CDAP 20826] Add an admission controller to inject init containers, node selectors and tolerations

arjan-bal:
## [CDAP 20826](https://cdap.atlassian.net/browse/CDAP-20826)

The PR adds a webhook server to the CDAP controller. The webhook server serves requests for a mutating admission controller which can modify pod specs. The configuration for modifying the pod specs is specified in the CDAPMaster CR. When a request to modify a pod spec comes, the webhook server first gets the corresponding CDAPMaster using the `cdap.instance` label in the pod. It then reads the `mutationConfigs` in the CDAPMaster to determine how to change the pod spec. For cases when the pod being deployed is not present in the same namespace as the CDAPMaster CR, a new label with key `cdap.k8s.namespace` will be added in CDAP.

Instructions to use the webhook and templates for using `cert-manager` for managing TLS certificates are added in the Readme file.


webhooks/cdap_webhook.go
```diff
@@ -0,0 +1,127 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+)
+
+const (
+	labelInstanceKey           = "cdap.instance"
+	// cdapMasterNamespaceKey is the label added by CDAP on pods launched by CDAP.
+	cdapMasterNamespaceKey     = "cdap.k8s.namespace"
+	// customResourceNamespaceKey is the label added by controller-reconciler on resources
+	// managed by the CDAP operator.
+	customResourceNamespaceKey = "custom-resource-namespace"
```
dli357, line 25: Is this the namespace of the CDAPMaster for a resource managed by the operator?

dli357 (approved):

webhooks/cdap_webhook.go
arjan-bal: Yes. The label will be added in CDAP 6.10 by https://github.com/cdapio/cdap/pull/15338. Without the label there isn't a way to identify the CDAPMaster for pods launched in other k8s namespaces.

arjan-bal: `custom-resource-namespace` is added to pods managed by the operator. It points to the namespace of the CDAP CR. Here is the responsible code: https://github.com/cdapio/cdap-operator/blob/develop/vendor_old/sigs.k8s.io/controller-reconciler/pkg/genericreconciler/handler.go#L62-L70

I've verified these labels are present in pods launched directly by the operator (not through twill).

PR #110, opened 2023-11-22
Fix mount permissions of secrets

rmstar:
Change mount permissions for secrets from 644 to 400.

NOTE: If pods are running as a non-root user, then [securityContext.fsGroup](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-the-security-context-for-a-pod) will need to be configured to ensure that the process running in the pod had access to secrets.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,7 +45,7 @@ manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and Cust
 
 .PHONY: generate
 generate: controller-gen ## Generate code containing DeepCopy, DeepCopyInto, and DeepCopyObject method implementations.
-	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./..."
+	#$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./..."
 
 .PHONY: fmt
 fmt: ## Run go fmt against code.
--- a/Makefile
+++ b/Makefile
@@ -45,7 +45,7 @@ manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and Cust
 
 .PHONY: generate
 generate: controller-gen ## Generate code containing DeepCopy, DeepCopyInto, and DeepCopyObject method implementations.
-	#$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./..."
+	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./..."
 
 .PHONY: fmt
 fmt: ## Run go fmt against code.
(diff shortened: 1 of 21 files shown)


albertshau:
could this break any existing instances?

rmstar:
> could this break any existing instances?

Yes, this can break instances where a pod running as a non-root user needs to access secrets. For example, the userinterface pods as as user `cdap` and reads [sConf](https://github.com/cdapio/cdap-ui/blob/develop/server.js#L144). The fix for this would be to configure [securityContext.fsGroup](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-the-security-context-for-a-pod) to ensure that the pod has access to secrets. 

AFAIK all other CDAP pods run as root, so they would not be affected.

albertshau:
what does it mean to configure the fsGroup for the UI pod? The UI pod isn't managed by the controller?

rmstar:
> what does it mean to configure the fsGroup for the UI pod? The UI pod isn't managed by the controller?

Yes, it's managed by the controller. The controller supports configuring fsGroup - https://github.com/cdapio/cdap-operator/blob/develop/api/v1alpha1/cdapmaster_types.go#L333.
You can set `spec.securityContext.fsGroup` in the CDAP CR.

For example, I have this set on my instance:
```
spec:
  securityContext:
    fsGroup: 1000
```

And the fsGroup settings are applied -
```
cdap@cdap-hdf683-userinterface-6fb7dc6bc7-2d6v5:/opt/cdap/ui$ id cdap
uid=1000(cdap) gid=1000(cdap) groups=1000(cdap)
cdap@cdap-hdf683-userinterface-6fb7dc6bc7-2d6v5:/opt/cdap/ui$ ls -lL /etc/cdap/security /etc/cdap/auth
/etc/cdap/auth:
total 8
-r--r----- 1 root cdap 136 Nov 27 17:41 cdap-key.json
-r--r----- 1 root cdap  13 Nov 27 17:41 realm.txt

/etc/cdap/security:
total 4
-r--r----- 1 root cdap 1300 Nov 27 17:41 cdap-security.xml
```

dli357:
Is it possible to set the group ownership for the mount path itself, or is that always configured to a fixed user (e.g. root) by Kubernetes? Otherwise, backfilling instances could be bothersome.

dli357 (approved):

rmstar:
> Is it possible to set the group ownership for the mount path itself, or is that always configured to a fixed user (e.g. root) by Kubernetes? Otherwise, backfilling instances could be bothersome.

Discussed offline. Mount path will be accessible by users that are in the configured fsGroup.

rmstar:
Closed in favor of #111

PR #111, opened 2023-12-08
Make mount permissions of secrets configurable

rmstar:
When default.mode.secret is set in CDAP config, the corresponding value (interpreted as decimal) will be used as the default mode for mounting CDAP secrets.


controllers/spec.go
```diff
@@ -147,6 +147,7 @@ type BaseSpec struct {
 	AdditionalVolumeMounts []corev1.VolumeMount      `json:"additionalVolumeMounts,omitempty"`
 	SecurityContext        *v1alpha1.SecurityContext `json:"securityContext,omitempty"`
 	Affinity               *corev1.Affinity          `json:"affinity,omitemtpy"`
+	DefaultModeSecret      int32                     `json:"defaultModeSecret,omitemtpy"`
```
dli357: I'm not entirely clear from the name what this refers to. Can we instead call this `SecretMountDefaultMode`? Or, even better, if this only applies to the SecuritySecret, we can move this below `SecuritySecretPath` and call it `SecuritySecretMountDefaultMode`

controllers/constants.go
```diff
@@ -64,13 +64,15 @@ const (
 	confTwillSecurityWorkerSecretDiskName = "twill.security.worker.secret.disk.name"
 	confTwillSecurityWorkerSecretDiskPath = "twill.security.worker.secret.disk.path"
 	confJMXServerPort                     = "jmx.metrics.collector.server.port"
+	confDefaultModeSecret                 = "default.mode.secret"
```
albertshau: nit: what about naming it secretDefaultMode and secret.default.mode instead

albertshau (approved):

controllers/spec.go
rmstar: Renamed to `SecretMountDefaultMode` as this sets the mount permissions for any secret, not just `cdap-security` managed by cdap.

controllers/constants.go
rmstar: Renamed - secret.mount.default.mode

PR #113, opened 2024-03-06
Support imagePullSecret for private registry.

mcameron:


mcameron:
Will just link the service account instead.

PR #138, opened 2025-02-12
Refactor upgrade logic to skip pre upgrade job in case of patch revisions

anshumanks:
### Description

This change prevents pre upgrade job from stopping pipelines in case of patch revisions.

### Code change

- Modified `version_update.go, constants.go`

### Unit Tests

- Modified `version_update_test.go`
- Modified old tests to incorporate new return value from compareVersions function (-2 for patch revisions)
- Added a new test to check for patch revisions

### Tested
In case of patch revisions:
Screenshot 2025-02-13 at 7 39 57 AM

Patch revision with skip pre upgrade flag = false:
Screenshot 2025-02-13 at 7 58 26 AM
Screenshot 2025-02-13 at 8 53 20 AM

In other cases:
Screenshot 2025-02-13 at 6 59 30 AM

### DIFF
--- a/controllers/constants.go
+++ b/controllers/constants.go
@@ -68,7 +68,7 @@ const (
 	confTwillSecurityWorkerSecretDiskPath = "twill.security.worker.secret.disk.path"
 	confJMXServerPort                     = "jmx.metrics.collector.server.port"
 	confSecretMountDefaultMode            = "secret.mount.default.mode"
-	confSkipPreUpgradeFlag                = "cdap-operator.preupgrade-job.skip"
+	confSkipPreUpgrade                    = "cdap-operator.preupgrade-job.skip"
 
 	// default values
 	defaultImage                  = "gcr.io/cdapio/cdap:latest"
--- a/controllers/constants.go
+++ b/controllers/constants.go
@@ -68,6 +68,7 @@ const (
 	confTwillSecurityWorkerSecretDiskPath = "twill.security.worker.secret.disk.path"
 	confJMXServerPort                     = "jmx.metrics.collector.server.port"
 	confSecretMountDefaultMode            = "secret.mount.default.mode"
+	confSkipPreUpgradeFlag                = "cdap-operator.preupgrade-job.skip"
 
 	// default values
 	defaultImage                  = "gcr.io/cdapio/cdap:latest"
--- a/controllers/version_update.go
+++ b/controllers/version_update.go
@@ -41,12 +41,12 @@ func handleVersionUpdate(master *v1alpha1.CDAPMaster, labels map[string]string, 
 		return nil, err
 	}
 	versionComparison := compareVersion(curVersion, newVersion)
-	isPatchRevision := versionComparison == -2
+	patchRevision := versionComparison == -4
 
 	// Let the current update complete if there is any
 	if isConditionTrue(master, updateStatus.Inprogress) {
 		log.Printf("Version update ingress. Continue... ")
-		return upgradeForBackend(master, labels, observed, isPatchRevision)
+		return upgradeForBackend(master, labels, observed, patchRevision)
 	}
 
 	if objs, versionUpdated, err := updateForUserInterface(master); err != nil {
@@ -62,7 +62,7 @@ func handleVersionUpdate(master *v1alpha1.CDAPMaster, labels map[string]string, 
 	}
 
 	switch versionComparison {
-	case -2, -1:
+	case -1, -2, -3, -4:
 		// Upgrade case
 
 		// Don't retry upgrade if it failed.
@@ -76,7 +76,7 @@ func handleVersionUpdate(master *v1alpha1.CDAPMaster, labels map[string]string, 
 		setCondition(master, updateStatus.Inprogress)
 		master.Status.UpgradeStartTimeMillis = getCurrentTimeMs()
 		log.Printf("Version update: start upgrading %s -> %s ", curVersion.rawString, newVersion.rawString)
-		return upgradeForBackend(master, labels, observed, isPatchRevision)
+		return upgradeForBackend(master, labels, observed, patchRevision)
 	case 0:
 		// Reset all condition so that failed upgraded/downgrade can be retried later if needed.
 		// This is needed when last upgrade failed and user has reset the version in spec.
@@ -123,7 +123,7 @@ func downgradeForBackend(master *v1alpha1.CDAPMaster) ([]reconciler.Object, erro
 	return []reconciler.Object{}, nil
 }
 
-func upgradeForBackend(master *v1alpha1.CDAPMaster, labels map[string]string, observed []reconciler.Object, isPatchRevision bool) ([]reconciler.Object, error) {
+func upgradeForBackend(master *v1alpha1.CDAPMaster, labels map[string]string, observed []reconciler.Object, patchRevision bool) ([]reconciler.Object, error) {
 	// Find either pre- or post- upgrade job
 	findJob := func(jobName string) *batchv1.Job {
 		var job *batchv1.Job = nil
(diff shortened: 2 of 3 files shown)


controllers/version_update.go
```diff
@@ -154,6 +157,19 @@ func upgradeForBackend(master *v1alpha1.CDAPMaster, labels map[string]string, ob
 		return jobObj
 	}
 
+	SkipPreUpgradeFlag := !(master.Spec.Config[confSkipPreUpgradeFlag] == "false")
```
vsethi09: nit: `skipPreUpgrade` instead of `SkipPreUpgradeFlag`

vsethi09: What happens when `confSkipPreUpgradeFlag` key is not present in `master.Spec.Config` map?

controllers/version_update.go
```diff
@@ -32,10 +32,21 @@ func init() {
...
+	curVersion, err := getCurrentImageVersion(master)
+	if err != nil {
+		return nil, err
+	}
+	newVersion, err := getNewImageVersion(master)
+	if err != nil {
+		return nil, err
+	}
+	versionComparison := compareVersion(curVersion, newVersion)
+	isPatchRevision := versionComparison == -2
```
vsethi09: nit: `patchRevision` instead of `isPatchRevision`

controllers/constants.go
```diff
@@ -68,6 +68,7 @@ const (
 	confTwillSecurityWorkerSecretDiskPath = "twill.security.worker.secret.disk.path"
 	confJMXServerPort                     = "jmx.metrics.collector.server.port"
 	confSecretMountDefaultMode            = "secret.mount.default.mode"
+	confSkipPreUpgradeFlag                = "cdap-operator.preupgrade-job.skip"
```
vsethi09: nit: simply name it `confSkipPreUpgrade`

controllers/version_update.go
```diff
@@ -154,6 +157,19 @@ func upgradeForBackend(master *v1alpha1.CDAPMaster, labels map[string]string, ob
...
 	}
 
+	SkipPreUpgradeFlag := !(master.Spec.Config[confSkipPreUpgradeFlag] == "false")
+
+	// Skip pre-upgrade and post-upgrade jobs for patch revision
+	if isPatchRevision && SkipPreUpgradeFlag {
+		log.Printf("Version update: patch revision detected, skipping pre-upgrade and post-upgrade jobs.")
+		setImageToUse(master)
+		setCondition(master, updateStatus.VersionUpdated)
+		setCondition(master, updateStatus.UpgradeSucceeded)
```
vsethi09: Data plane still needs to be updated. Why is `setCondition(master, updateStatus.UpgradeSucceeded)` used here?

controllers/version_update.go
```diff
@@ -154,6 +157,19 @@ func upgradeForBackend(master *v1alpha1.CDAPMaster, labels map[string]string, ob
 		return jobObj
 	}
 
+	SkipPreUpgradeFlag := !(master.Spec.Config[confSkipPreUpgradeFlag] == "false")
+
+	// Skip pre-upgrade and post-upgrade jobs for patch revision
+	if isPatchRevision && SkipPreUpgradeFlag {
+		log.Printf("Version update: patch revision detected, skipping pre-upgrade and post-upgrade jobs.")
+		setImageToUse(master)
```
vsethi09: This code is redundant with line 208 - 214

controllers/version_update.go
```diff
@@ -418,9 +435,24 @@ func compareVersion(l, r *Version) int {
 		return -1
 	}
 
+	lenL, lenR := len(l.components), len(r.components)
+	// Check if it only a patch revision
```
vsethi09: nit: fix comment

controllers/version_update.go
```diff
@@ -406,6 +422,7 @@ func parseImageString(imageString string) (*Version, error) {
 }
 
 // compare two parsed versions
+// -2: left < right, patch revision
```
vsethi09: For simplicity return the index of the component which does not match.

Return <0 values if left < right
Return 0 if left == right
Return 1 if left > right

controllers/version_update.go
```diff
@@ -418,9 +435,24 @@ func compareVersion(l, r *Version) int {
 		return -1
 	}
 
+	lenL, lenR := len(l.components), len(r.components)
+	// Check if it only a patch revision
+	if lenL == lenR && lenL > 0 && l.components[lenL-1] < r.components[lenL-1] {
```
vsethi09: Checking for `lenL > 0` may not be sufficient.

What happens when left = 6.11.0.1 and right = 6.11.0?

controllers/version_update.go
anshumanks: In that case it is null, i.e., not false. So unless we set it explicitly to false, skipPreUpgrade will remain "true".
Tested it without confSkipPreUpgradeFlag in the config, and then added it there. Behavior was as expected.

controllers/version_update.go
anshumanks: Now returning -1/-2/-3/-4 based on where the difference is when left < right.

controllers/version_update.go
anshumanks: With the change in compareVersions(-n when left < right), this is handled now.

controllers/version_update.go
anshumanks: PTAL at the changes. Now simply skipping pre and post upgrade jobs instead of doing anything in this block.

controllers/version_update.go
anshumanks: PTAL at the changes. Now simply skipping pre and post upgrade jobs instead of doing anything in this block.

controllers/version_update.go
```diff
@@ -406,7 +419,7 @@ func parseImageString(imageString string) (*Version, error) {
 }
 
 // compare two parsed versions
-// -1: left < right
+// -n: left < right, nth component differs (1-indexed)
 // 0: left = right
 // 1: left > right
 func compareVersion(l, r *Version) int {
```
vsethi09, line 428: `compareVersion()` can be simplified to similarly return 1,2,3,4 when left > right. Moreover, you won't need to run 3 for loops. The result can be obtained in one for loop.

The caller doesn't need to use `switch case`. It can instead use `if..else if...else` construct.

controllers/version_update.go
```diff
@@ -154,6 +157,16 @@ func upgradeForBackend(master *v1alpha1.CDAPMaster, labels map[string]string, ob
 		return jobObj
 	}
 
+	// Skip pre-upgrade and post-upgrade jobs for patch revisions
+	skipPreUpgrade := patchRevision && !(master.Spec.Config[confSkipPreUpgrade] == "false")
+	if skipPreUpgrade {
+		log.Printf("Version update: patch revision detected, skipping pre-upgrade and post-upgrade jobs.")
+		// Mark pre and post upgrade jobs as succeeded so they don't get triggered when reconciling
+		// patchRevision will become false after reconciliation
+		setCondition(master, updateStatus.PreUpgradeSucceeded)
```
vsethi09: No need to mark them as succeeded if you haven't run it.

In case of patch revision, the logic would probably the same as in the case of `versionComparison == 0`.

controllers/version_update.go
```diff
@@ -154,6 +157,16 @@ func upgradeForBackend(master *v1alpha1.CDAPMaster, labels map[string]string, ob
 		return jobObj
 	}
 
+	// Skip pre-upgrade and post-upgrade jobs for patch revisions
```
vsethi09: All the logic to check whether Pre Upgrade job should run should be done at the start. Line  127.

controllers/version_update.go
anshumanks: That was the initial plan, but if I add something like `if !isConditionTrue(master, updateStatus.Pre/PostUpgradeSucceeded) && !skipPreUpgrade`, this will skip pre upgrade, and change the image. Then versionComparison will become 0. 
If we incorporate 0 along with -4 in skipPreUpgrade ( `versionComparison == -4 || versionComparison == 0` ), for minor/major changes the post upgrade job will get skipped.

Possible fix: Skip pre upgrade in case of 0 and -4, and modify post upgrade condition as 
`!isConditionTrue(master, updateStatus.PostUpgradeSucceeded) && isConditionTrue(master, updateStatus.PreUpgradeSucceeded)`

controllers/version_update.go
```diff
@@ -45,23 +56,12 @@ func handleVersionUpdate(master *v1alpha1.CDAPMaster, labels map[string]string,
...
 	if len(curVersion.rawString) == 0 {
 		setImageToUse(master)
 		return []reconciler.Object{}, nil
 	}
 
-	switch compareVersion(curVersion, newVersion) {
-	case -1:
-		// Upgrade case
-
+	if versionComparison < 0 { // Upgrade case
```
vsethi09: nit: please add comments above the code instead of inline. Comment sentences should end with a period (`.`)

Fix t/o

vsethi09 (approved):

PR #143, opened 2025-03-06
[CDAP-21153] Run StorageMain for appfabric deployment pod as initContainer

vsethi09:
[CDAP-21153] Run StorageMain for appfabric deployment pod as initContainer

### Issue
CDAP instance upgraded to 6.11 may have errors like `io.cdap.cdap.spi.data.InvalidFieldException: Field flow_control_status is not part of the schema of table run_records`.

### Root Cause

This happens when there is a race condition where Appfabric service starts up and loads schema (caches) from DB before other services initialize storage table with new fields. The loaded schema doesn't contain the new field and fails to write run record during execution of pipeline.

### Impact

If appfabric starts before the schema is updated after instance upgrade all the new tables and fields introduced will be missing in the schema cache, which can cause read / write operations to fail. This can happen for instance upgrade from any older version (eg. 6.9 where SCM was introduced with new tables and fields or 6.8 where new fields were added for LCM)

### Solution

Run storage init as part of appfabric pod, like other statefulsets. 

### Verification

- [x] Unit Tests
- [x] Docker image

[CDAP-21096]: https://cdap.atlassian.net/browse/CDAP-21096?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[CDAP-21153]: https://cdap.atlassian.net/browse/CDAP-21153?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/controllers/deployment.go
+++ b/controllers/deployment.go
@@ -409,6 +409,14 @@ func buildDeployment(master *v1alpha1.CDAPMaster, name string, services ServiceG
 		setAffinity(affinity).
 		setSecretMountDefaultMode(defaultMode)
 
+	// Run storage init for appfabic deployment pod.
+	// TODO(CDAP-21152): Either run storage init once during upgrade or enable
+	// for more deployments.
+	if name == "appfabric" {
+		spec = spec.withInitContainer(
+			newContainerSpec(master, "StorageInit", dataDir).setArgs(containerStorageMain))
+	}
+
 	// Add each service as a container
 	for _, s := range services {
 		ss, err := getCDAPServiceSpec(master, s)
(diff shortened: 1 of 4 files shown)


itsankit-google (commented):
please add a jira to PR title

sidhdirenge (approved):

itsankit-google (approved):

